PR #19, opened 2021-05-11
Ownerless DBC's with a singleton Mint are feature complete

davidrusu:
This PR adds a few missing validations:
* ensure outputs are numbered
* ensure output parents are the inputs

Added the fuzz testing for these two validations as well.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -40,6 +40,7 @@ fn sha3_256(input: &[u8]) -> Hash {
 mod tests {
     use super::*;
 
+    use core::num::NonZeroU8;
     use quickcheck::{Arbitrary, Gen};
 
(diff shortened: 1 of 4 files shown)


src/lib.rs
```diff
@@ -67,12 +67,37 @@ mod tests {
         }
     }
 
+    #[derive(Clone, Copy, PartialEq, Eq, PartialOrd, Ord)]
+    pub struct NonZeroTinyInt(u8);
```
Dylan-DPC, line 71: maybe use [`NonZeroU8`](https://doc.rust-lang.org/std/num/struct.NonZeroU8.html) ?

davidrusu: :+1: Ok pushed up a change that makes use of the NonZeroU8 type, I think it's better now

dirvine (approved):

PR #32, opened 2021-05-22
feat(mint): take input hashes belonging to mint, in reissue method

oetyng:
- Add custom error for validation that the filtered input actually is
present in the transaction as well.


davidrusu (approved):
:shark: lgtm

PR #33, opened 2021-05-23
Hash pr

dan-da:
These are some mods I made to enable displaying mint and dbc information nicely in a CLI.

The largest part of the PR changes `Hash` to use newtype pattern instead of a type alias.  I initially did this in order to impl fmt::Display and Debug on Hash for pretty printing using base64 encoding.  However, I discovered that fmt::Display really cannot do what I need, so in the end the app does all the display formatting itself.  I also reverted the custom base64 encoding in Hash::Debug, as I figure it is perhaps better to display the raw data/structure in Debug mode.  As such, the newtype pattern could be reverted from the PR, but I think it is useful to keep, as it makes it easy/possible to add functionality to Hash and DbcContentHash in the future.

Other changes are:
* adds PublicKey::to_bytes()
* derive Debug for Mint, KeyCache, KeyManager, SpendBook

These last two are relied upon by the cli mint prototype.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -7,11 +7,14 @@
 // permissions and limitations relating to use of the SAFE Network Software.
 #![allow(clippy::from_iter_instead_of_collect)]
 
+use serde::{Deserialize, Serialize};
+use std::ops::Deref;
 #[cfg(test)]
 use tiny_keccak::{Hasher, Sha3};
 /// These typdefs are to simplify algorithm for now and will be removed for production.
-pub(crate) type Hash = [u8; 32];
-pub(crate) type DbcContentHash = [u8; 32];
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Hash, PartialOrd, Ord, Serialize, Deserialize)]
+pub struct Hash([u8; 32]);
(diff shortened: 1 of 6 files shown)


src/lib.rs
```diff
@@ -28,6 +33,32 @@ pub use crate::{
...
+    fn from(val: [u8; 32]) -> Hash {
+        Hash(val)
+    }
+}
+
+impl Distribution<Hash> for Standard {
+    fn sample<R: Rng + ?Sized>(&self, rng: &mut R) -> Hash {
+        Hash(rng.gen())
+    }
+}
```
dirvine, line 46: This seems a bit weird, is it required? (apart from that, great addition to view Dbcs like this)

davidrusu: It's only used when fuzzing DBC's in testing.

@dan-da lets add a `#[cfg(test)]` in front of this impl as to not pollute the public API

dan-da: k, will do

dan-da: done.

davidrusu (dismissed):

davidrusu (approved):

PR #41, opened 2021-05-31
feat(mint): replace ed25519 mint identities with BLS

davidrusu:


src/dbc.rs
```diff
@@ -153,8 +153,29 @@ mod tests {
     ) {
         let amount = 100;
         let genesis_owner = crate::bls_dkg_id();
-        let (mut genesis, genesis_dbc) =
-            Mint::genesis(genesis_owner.public_key_set.clone(), amount);
+        let genesis_key = genesis_owner.public_key_set.public_key();
```
dan-da, line 156: This is gonna change to be PublicKeySet not PublicKey, right?

I guess I will hold off reviewing further, in case there is an update coming.

dan-da (dismissed):

dan-da (approved):

PR #50, opened 2021-06-09
refactor(signing): impl async for signer trait

oetyng:
This is because downstream consumer dependency implements async methods returning result, for signing.

### DIFF
--- a/src/dbc.rs
+++ b/src/dbc.rs
@@ -62,7 +62,7 @@ mod tests {
     use quickcheck_macros::quickcheck;
 
     use crate::tests::{NonZeroTinyInt, TinyInt};
-    use crate::{KeyManager, Mint, ReissueRequest, ReissueTransaction};
+    use crate::{ExposedSigner, KeyManager, Mint, ReissueRequest, ReissueTransaction};
 
     fn divide(amount: u64, n_ways: u8) -> impl Iterator<Item = u64> {
         (0..n_ways).into_iter().map(move |i| {
@@ -156,8 +156,11 @@ mod tests {
         let genesis_key = genesis_owner.public_key_set.public_key();
 
         let mut genesis_node = Mint::new(KeyManager::new(
-            genesis_owner.public_key_set.clone(),
-            (0, genesis_owner.secret_key_share.clone()),
+            ExposedSigner::new(
+                0,
+                genesis_owner.public_key_set.clone(),
+                genesis_owner.secret_key_share.clone(),
+            ),
             genesis_key,
         ));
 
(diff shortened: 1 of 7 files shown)


src/key_manager.rs
```diff
@@ -53,33 +53,64 @@ impl From<Vec<PublicKey>> for KeyCache {
...
 
+pub trait Signer {
+    fn index(&self) -> u64;
+    fn pk_set(&self) -> PublicKeySet;
+    fn sign<M: AsRef<[u8]>>(&self, msg: M) -> SignatureShare;
+}
+
+pub struct ExposedSigner {
+    index: u64,
+    pk_set: PublicKeySet,
```
dan-da, line 66: Can we call this public_key_set for clarity and symmetry with secret_key_set?

src/key_manager.rs
```diff
@@ -53,33 +53,64 @@ impl From<Vec<PublicKey>> for KeyCache {
...
+        }
+    }
+}
+
+impl Signer for ExposedSigner {
+    fn index(&self) -> u64 {
+        self.index
+    }
+
+    fn pk_set(&self) -> PublicKeySet {
```
dan-da: likewise public_key_set() here.       "pk" could also mean private key.

src/key_manager.rs
```diff
@@ -53,33 +53,64 @@ impl From<Vec<PublicKey>> for KeyCache {
     }
 }
 
+pub trait Signer {
+    fn index(&self) -> u64;
+    fn pk_set(&self) -> PublicKeySet;
```
dan-da: and here.  (see below comment)

src/dbc.rs
```diff
@@ -155,14 +154,20 @@ mod tests {
         let genesis_owner = crate::bls_dkg_id();
         let genesis_key = genesis_owner.public_key_set.public_key();
 
-        let mut genesis_node = Mint::new(KeyManager::new(
-            genesis_owner.public_key_set.clone(),
-            (0, genesis_owner.secret_key_share.clone()),
-            genesis_key,
-        ));
+        let mut genesis_node = Mint::new(
+            block(KeyManager::new(
```
davidrusu, line 158: You sure this needs a `block`?

src/dbc.rs
```diff
@@ -297,12 +300,12 @@ mod tests {
         for _ in 0..n_wrong_signer_sigs.coerce() {
             if let Some(input) = repeating_inputs.next() {
                 let id = crate::bls_dkg_id();
-                let key_mgr = KeyManager::new(
-                    id.public_key_set.clone(),
-                    (0, id.secret_key_share),
+                let key_mgr = block(KeyManager::new(
```
davidrusu, line 303: again, I don't think this needs a `block`

dan-da:
This PR is titled "use signer trait instead of secret key share".    But it is now introducing async.

I'd prefer to see them separated into 2 distinct PRs if possible...?

oetyng:
> This PR is titled "use signer trait instead of secret key share". But it is now introducing async.
> 
> I'd prefer to see them separated into 2 distinct PRs if possible...?

Sure, first one is here: https://github.com/maidsafe/sn_dbc/pull/51

oetyng:
Superseded by #61

PR #51, opened 2021-06-09
refactor(signing): use signer trait instead of secret key share

oetyng:
This is because downstream consumer does not have access to secret key shares, and instead relies on a dependency to hold it and sign data for them.

### DIFF
--- a/src/key_manager.rs
+++ b/src/key_manager.rs
@@ -53,33 +53,64 @@ impl From<Vec<PublicKey>> for KeyCache {
     }
 }
 
-#[derive(Debug, Clone)]
-pub struct KeyManager {
+pub trait Signer {
+    fn index(&self) -> u64;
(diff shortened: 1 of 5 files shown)


src/key_manager.rs
```diff
@@ -91,17 +122,16 @@ impl KeyManager {
...
-        NodeSignature(
-            self.secret_key_share.0,
-            self.secret_key_share.1.sign(msg_hash),
-        )
+        NodeSignature(self.signer.index(), self.signer.sign(msg_hash))
     }
 
     pub fn verify(&self, msg_hash: &Hash, key: &PublicKey, signature: &Signature) -> Result<()> {
+        // NB: this can fail if self.signer changes keys..
+        // then cache needs to be kept in sync with signer view of its current key
```
davidrusu, line 134: When would a signer change keys?

oetyng:
Superseded by #60

PR #56, opened 2021-06-15
add mint-repl and associated readme, docs

dan-da:
A note about naming:

I was just going to name this example "mint".

However, In order to keep the code simpler and contained in a single file, I've decided to implement cli arguments feature as a separate example.  As such, we will end up with mint-cli example and mint-repl example.  The mint-cli example will require persistence to disk, which the repl does not.


examples/mint-repl/mint-repl.rs
```diff
@@ -0,0 +1,1110 @@
...
+    fn mintnode(&self) -> Result<&Mint> {
+        self.mintnodes
+            .get(0)
+            .ok_or_else(|| anyhow!("Mint not yet created"))
+    }
+}
+
+/// A Dbc plus the owner's pubkey set
+#[derive(Debug, Clone, PartialEq, Eq, Hash, Serialize, Deserialize)]
+struct DbcOwned {
```
davidrusu, line 49: Perhaps `DbcUnblinded`?

examples/mint-repl/mint-repl.rs
```diff
@@ -0,0 +1,1110 @@
...
+    let mut rl = Editor::<()>::new();
+    rl.set_auto_add_history(true);
+    'outer: loop {
+        match rl.readline(">> ") {
+            Ok(line) => {
+                let mut args = line.trim().split_whitespace();
+                let cmd = if let Some(cmd) = args.next() {
+                    cmd
+                } else {
+                    continue 'outer;
```
davidrusu, line 90: Is this `outer` label still required? I'm not seeing an inner loop here

examples/mint-repl/mint-repl.rs
```diff
@@ -0,0 +1,1110 @@
...
+fn newmint() -> Result<MintInfo> {
+    let confirm = readline_prompt(
+        "\nThis will erase existing Mint and transactions.  Are you sure? [y/n]: ",
+    )?;
+    if confirm != "y" {
+        return Err(anyhow!("newmint operation cancelled"));
+    }
+
+    let amount = loop {
+        let amount: u64 = readline_prompt("\nTotal Money Supply Amount: ")?.parse()?;
```
davidrusu, line 139:
```suggestion
        let amount: u64 = readline_prompt("\nTotal Money Supply Amount (> 0): ")?.parse()?;
```

examples/mint-repl/mint-repl.rs
```diff
@@ -0,0 +1,1110 @@
...
+        }
+    };
+
+    // polynomial, from which SecretKeySet is built.
+    let poly_input = readline_prompt_nl("\nSecretKeySet Poly Hex, or [r]andom: ")?;
+
+    let mintinfo = match poly_input.as_str() {
+        "r" => {
+            let threshold = loop {
+                let num_signers: usize = readline_prompt("\nHow many signers: ")?.parse()?;
```
davidrusu, line 152:
```suggestion
                let num_signers: usize = readline_prompt("\nHow many signers (> 0): ")?.parse()?;
```

examples/mint-repl/mint-repl.rs
```diff
@@ -0,0 +1,1110 @@
...
+    println!("amount: {}\n", dbc.inner.content.amount);
+    println!("output_number: {}\n", dbc.inner.content.output_number);
+    println!("owner: {}\n", to_be_hex(&dbc.owner)?);
+
+    // dbc.content.parents and dbc.transaction.inputs seem to be the same
+    // so for now we are just displaying the latter.
+    // println!("parents:");
+    // for p in &dbc.content.parents {
+    //     println!("  {}", encode(p))
+    // }
```
davidrusu, line 371: Yep, they are the same for valid DBC's. One of the validation steps is checking that transaction inputs matches content inputs.

examples/mint-repl/mint-repl.rs
```diff
@@ -0,0 +1,1110 @@
...
+        r#"
+ __     _                        
+(_  _._|__  |\ | __|_     _ ._|  
+__)(_| |(/_ | \|(/_|_\/\/(_)| |< 
+ ____  ____   ____   __  __ _       _   
+|  _ \| __ ) / ___| |  \/  (_)_ __ | |_ 
+| | | |  _ \| |     | |\/| | | '_ \| __|
+| |_| | |_) | |___  | |  | | | | | | |_ 
+|____/|____/ \____| |_|  |_|_|_| |_|\__|        
+  "#
```
davidrusu, line 473: nice

examples/mint-repl/mint-repl.rs
dan-da: I was thinking maybe I should just get rid of the > 0 check, and let sn_dbc validation logic handle it.  Especially in light of recent discussions that issuing 0 amounts Dbcs is presently possible.

examples/mint-repl/mint-repl.rs
dan-da: yeah, makes sense.

examples/mint-repl/mint-repl.rs
```diff
@@ -0,0 +1,1110 @@
...
+/// Deserialize anything deserializable from big endian bytes
+fn from_be_bytes<T: for<'de> Deserialize<'de>>(b: &[u8]) -> Result<T> {
+    let bb = big_endian_bytes_to_bincode_bytes(b.to_vec());
+    bincode::deserialize(&bb).map_err(|e| anyhow!(e))
+}
+
+/// Deserialize anything deserializable from big endian bytes, hex encoded.
+fn from_be_hex<T: for<'de> Deserialize<'de>>(s: &str) -> Result<T> {
+    from_be_bytes(&decode(s)?)
+}
```
davidrusu, line 1046: These could be useful for sn_client once DBC's are exposed there, need to remember they are here.

davidrusu (dismissed):
I like the sample runs, nice touch.

Looks good!

examples/mint-repl/mint-repl.rs
dan-da: good catch.  that came from qp2p example I started with, and I might have removed a loop somewhere in the process.  I'm a little surprise clippy doesn't check for that.   anyway, I'll change it.

examples/mint-repl/mint-repl.rs
davidrusu: Yeah, that sounds better actually. Don't shield the DBC validation logic

dan-da: on second thought, I've changed it to a warning/prompt, and let user proceed or re-enter.

examples/mint-repl/mint-repl.rs
dan-da: yeah very useful since threshold_crypto api avoids exposing raw bytes.  But really any non-trivial app will require them.

dan-da:
@davidrusu I made the suggested changes.    In particular, amount validation is gone, although it does warn if amount is 0 or if sum(outputs) is greater than sum(inputs).   I verified that sn_dbc give an error for this case, and also that it issue 0 amount certs.

davidrusu (approved):

PR #60, opened 2021-06-18
refactor(signing): use keymanager trait instead of exposing secret ke…

oetyng:
…y share
Required by downstream sn consumer.

### DIFF
--- a/examples/mint-repl/mint-repl.rs
+++ b/examples/mint-repl/mint-repl.rs
@@ -18,11 +18,8 @@ use sn_dbc::{
     BlindedOwner, Dbc, DbcContent, DbcTransaction, Hash, Mint, MintSignatures, NodeSignature,
     ReissueRequest, ReissueTransaction, SimpleKeyManager as KeyManager, SimpleSigner as Signer,
 };
+use std::collections::{BTreeMap, BTreeSet, HashMap, HashSet};
 use std::iter::FromIterator;
-use std::{
-    collections::{BTreeMap, BTreeSet, HashMap, HashSet},
-    sync::Arc,
-};
 use threshold_crypto::poly::Poly;
 use threshold_crypto::serde_impl::SerdeSecret;
 use threshold_crypto::{
@@ -194,13 +191,13 @@ fn mk_new_mint(secret_key_set: SecretKeySet, poly: Poly, amount: u64) -> Result<
     let mut genesis_set: Vec<(DbcContent, DbcTransaction, (PublicKeySet, NodeSignature))> =
         Default::default();
     for i in 0..secret_key_set.threshold() as u64 + 1 {
-        let key_manager = Arc::new(KeyManager::new(
+        let key_manager = KeyManager::new(
             Signer::new(
                 secret_key_set.public_keys().clone(),
                 (i, secret_key_set.secret_key_share(i).clone()),
             ),
             genesis_pubkey,
-        ));
+        );
         let mut mint = Mint::new(key_manager);
         genesis_set.push(mint.issue_genesis_dbc(amount)?);
         mints.push(mint);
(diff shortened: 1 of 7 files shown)


examples/mint-repl/mint-repl.rs
```diff
@@ -185,17 +188,19 @@ fn mk_new_random_mint(threshold: usize, amount: u64) -> Result<MintInfo> {
...
+    let mut mints: Vec<Mint<KeyManager>> = Default::default();
 
     // Generate each Mint node, and corresponding NodeSignature. (Index + SignatureShare)
     let mut genesis_set: Vec<(DbcContent, DbcTransaction, (PublicKeySet, NodeSignature))> =
         Default::default();
     for i in 0..secret_key_set.threshold() as u64 + 1 {
-        let key_manager = KeyManager::new(
-            secret_key_set.public_keys().clone(),
-            (i, secret_key_set.secret_key_share(i).clone()),
+        let key_manager = Arc::new(KeyManager::new(
```
dan-da, line 197: why does this need to use Arc?

oetyng: Due to line 189 in `mint.rs`:

```rust
    pub fn verifier(&self) -> Verifier<K> {
        Verifier::new(self.key_manager.clone())
    }
```

dan-da: ok, I see there is:

```
pub struct Verifier<K: KeyManager> {
    key_manager: Arc<K>,
}

impl<K: KeyManager> Verifier<K> {
    pub fn new(key_manager: Arc<K>) -> Self {
        Self { key_manager }
    }
```

I just don't understand yet the reason for (a) using Arc in sn_dbc, and (b) exposing/requiring it at API level, ie Verifier::new(km: Arc<K>)

I guess I'm just wondering if we are mixing concerns here.  Maybe this is anticipating changes in the async PR?

oetyng: It's to be able to return a `Verifyer` from the `Mint`., as to pass in something with a limited API to the dbc when confirming valid.
That was the original design, and in the discussion it was said to be better not to pass in `KeyManager` there.

Now that we don't actually hold secret keys, we only hold a reference to something that updates under the hood.
Unless we want to use that thing directly this is sort of the thing I'd say, as we can't have a verifier with a reference to keymanager without it. At least this is a simple way.

But I can just pass in the `KeyManager` straight away, if exposing that fulll api within DBC is less of a problem than Arc.

oetyng: Pushed a commit with that approach now.

src/dbc.rs
```diff
@@ -30,13 +31,15 @@ impl Dbc {
 
     // Check there exists a DbcTransaction with the output containing this Dbc
     // Check there DOES NOT exist a DbcTransaction with this Dbc as parent (already minted)
-    pub fn confirm_valid(&self, key_cache: &KeyCache) -> Result<(), Error> {
+    pub fn confirm_valid<K: KeyManager>(&self, verifier: &K) -> Result<(), Error> {
```
davidrusu, line 34: Another approach would be to have `V: Verifier` here and split up the `KeyManager` trait into two traits:

```rust
trait DbcSigner {
    ... signing functions
}
trait DbcVerifier {
    fn verify(..) {}
}
impl DbcSigner for SimpleKeyManager {
    ...
}
impl  DbcVerifier for SimpleKeyManager {
    ...
}
```
That way you can expose only the verifier api to the Dbc::confirm_valid despite using the same struct behind the scenes for signing as well.

src/key_manager.rs
```diff
@@ -13,99 +13,133 @@ use threshold_crypto::{serde_impl::SerdeSecret, SecretKeyShare, SignatureShare};
 pub use threshold_crypto::{PublicKey, PublicKeySet, Signature};
 
 #[derive(Debug, Clone, Hash, PartialEq, Eq, Deserialize, Serialize)]
-pub struct NodeSignature(u64, SignatureShare);
+pub struct NodeSignature {
+    index: u64,
+    sig: SignatureShare,
+}
```
davidrusu, line 19: nice :)

davidrusu (approved):
Looks good to me :ship:

PR #63, opened 2021-06-23
make SpendBook a trait so that implementer can decide how to store it

dan-da:
first cut at this.

I made IntoIterator a supertrait of SpendBook trait to provide/require a mechanism for reading all the spendbook entries.   I'm not sure if there is any way to enforce that implementer's actually return a DbcTransaction when iterating though... seems they could return any type they wish.  any suggestions?

Also, I wanted to name the default SpendBook impl `SpendBookRam` as it seems more descriptive.  But I went with `SimpleSpendBook` for symmetry with the other traits.

Also, I snuck in a tiny fix for mint-repl `newmint` command.

### DIFF
--- a/src/mint.rs
+++ b/src/mint.rs
@@ -28,8 +28,14 @@ pub type MintSignatures = BTreeMap<DbcContentHash, (PublicKeySet, NodeSignature)
 pub const GENESIS_DBC_INPUT: Hash = Hash([0u8; 32]);
 
 pub trait SpendBook: std::fmt::Debug + Clone + IntoIterator {
-    fn lookup(&self, dbc_hash: &DbcContentHash) -> Option<&DbcTransaction>;
-    fn log(&mut self, dbc_hash: DbcContentHash, transaction: DbcTransaction);
+    type Error: std::error::Error;
+
+    fn lookup(&self, dbc_hash: &DbcContentHash) -> Result<Option<&DbcTransaction>, Self::Error>;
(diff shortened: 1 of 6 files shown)


src/mint.rs
```diff
@@ -27,12 +27,17 @@ pub type MintSignatures = BTreeMap<DbcContentHash, (PublicKeySet, NodeSignature)
 
 pub const GENESIS_DBC_INPUT: Hash = Hash([0u8; 32]);
 
+pub trait SpendBook: std::fmt::Debug + Clone + IntoIterator {
+    fn lookup(&self, dbc_hash: &DbcContentHash) -> Option<&DbcTransaction>;
+    fn log(&mut self, dbc_hash: DbcContentHash, transaction: DbcTransaction);
```
davidrusu, line 32: Lets have these methods return a `Result<_, Self::Error>` and use an associated `Error` type on the trait. As soon as we start doing IO in the spendbook, we'll have to return Results.

(You can use `Error=Infallible` for the `SimpleSpendbook` impl)

dan-da: good idea, will do.

dan-da: ok, I impl'd this following the same pattern as was used for KeyManager.   So there is an Error::SpendBook(String) and I call map_err() to convert from the trait error.

davidrusu (approved):
Looks good :ship:

PR #69, opened 2021-07-08
add mint-cli example

dan-da:
I created this example in response to request from @dirvine and @oetyng some time ago when we were discussing adding the repl example and they wanted it to be scriptable via the CLI. 

It turns out that this added a lot of complexity/code to the repl example, so I broke them out into separate examples.

One thing I am unhappy about with this PR is that it adds 4 new dev-deps.   Unfortunately cargo does not yet allow specifying deps per artifact, eg per example.  Though [rfc 2887](https://github.com/rust-lang/rfcs/pull/2887) would enable that, so maybe one day.

If it is judged that the extra dev-deps are not acceptable, I can just maintain it in a separate repo.  Or we could consider making it a sub-crate, though that seems to me like overkill for a library example.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -42,10 +42,16 @@ rand = "0.7.1"
 bincode = "1.3.3"
 xor_name = "1.2.0"
 hex = "0.4.3"
+structopt = "0.3.21"
+serde_derive = "1.0.126"
+preferences-ron = "2.0.0"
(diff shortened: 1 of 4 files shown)


davidrusu (approved):
:ship:

dan-da:
cleanup:  closing out-dated PRs.

After sn_dbc api has stabilized I might make a separate crate for the mint-cli example.  Or maybe just the dbc playground instead.

PR #72, opened 2021-08-03
confidential transactions.  pedersen commitments + bulletproofs (range proofs)

dan-da:
See PR #70 for previous discussion.

Here I have rebased everything into a single commit.


davidrusu:
Some benchmarks comparing confidential DBC's with those in master.

As expected, the ZK math is more expensive, merging is not much worse, but splitting is quite a lot slower.
### master
#### Merge 100 DBC's into 1
Screen Shot 2021-08-09 at 12 58 59 AM
#### Split 1 DBC into 100 DBC's
Screen Shot 2021-08-09 at 12 59 53 AM

### confidential_amounts_pr
#### Merge 100 DBC's into 1
Screen Shot 2021-08-09 at 1 07 16 AM

#### Split 1 DBC into 100 DBC's
Screen Shot 2021-08-09 at 1 07 57 AM

dan-da:
thx @davidrusu!    So just to summarize then:

merging:
* original:  ~150.8ms
* confidential: ~166.3ms  (1.1x)
 
splitting:
* original: ~2.63ms
* confidential: ~215.65 (~82x)

dan-da:
@davidrusu can you clarify if the benches only for mint processing or they also include client processing?  

In particular if the benches include DbcContent::new() then that means they are counting the time to generate the range proof (by the client).  Verification time is asymmetric with generation time, so the mint should have less work to do...

davidrusu:
@dan-da Benchmarks are only covering the mint code path (`Mint::reissue`)

dan-da:
btw, the discrepancy between the split and merge times makes sense to me because Mint::validate_balances() is verifying the range proof only for outputs.    My logic is that the inputs were already verified by the mint at time of creation and carry the mint's signature.  So long as the sig is good, we don't need to revalidate.

So a merge is creating only a single output, while a split is creating 100.  Hence the split will verify proofs 99 more times.

Cargo.toml
```diff
@@ -33,15 +37,21 @@ hex = "0.4.3"
   version = "1.0.111"
   features = [ "derive", "rc" ]
 
+[patch.crates-io]
+# Use this branch until it is merged into bls_dkg master.
+bls_dkg = { git = "https://github.com/maidsafe/bls_dkg", branch="blsttc_2_3_0_pr" }
```
davidrusu, line 42: https://github.com/maidsafe/bls_dkg/pull/94 is merged now

benches/reissue.rs
```diff
@@ -1,15 +1,27 @@
...
 };
 
 use criterion::{black_box, criterion_group, criterion_main, Criterion};
 
+fn decrypt_amount_secrets(
+    owner: &bls_dkg::outcome::Outcome,
+    dbcc: &DbcContent,
+) -> Result<AmountSecrets, Error> {
+    let mut shares: BTreeMap<usize, SecretKeyShare> = Default::default();
+    shares.insert(0, owner.secret_key_share.clone());
```
davidrusu, line 20:
```suggestion
    let shares = BTreeMap::from_iter([(0, owner.secret_key_share.clone())]);
```

davidrusu (dismissed):
🚢 nice work.

With this change, sn_dbc users need to do a lot of work to use this library properly now. I think we are in a more real need of a builder api to help with things like making sure the blinding factors are generated properly and balance etc.

Perhaps we have unblinded versions of DbcContent that just stores the intention, then we send the set of unblinded dbc contents to the builder that then creates a ReissueRequest? Something along those lines.

We can do that in a next PR as this is already large.

Cargo.toml
dan-da: cool, using published version now.

benches/reissue.rs
dan-da: done.

davidrusu (approved):

PR #74, opened 2021-08-13
Transaction Builder

davidrusu:
Changes:
- transaction builder to simplify usage
- `from_iter(vec![..])` to `from_iter([..]`
- remove `DbcContent.output_number`, it was added to differentiate between identical outputs, but we have blinding factors in Pederson commitments doing that job now.


examples/mint-repl/mint-repl.rs
```diff
@@ -1132,7 +1128,7 @@ fn reissue_exec(
         .collect();
 
     // sort outputs by output_number
```
dan-da, line 1130:
```suggestion
    // sort outputs by name
```

src/builder.rs
```diff
@@ -0,0 +1,81 @@
+use std::collections::{BTreeSet, HashMap, HashSet};
+use std::iter::FromIterator;
+
+use curve25519_dalek_ng::scalar::Scalar;
+
+use crate::{AmountSecrets, Dbc, DbcContent, ReissueTransaction, Result};
+
+///! Unblinded data for creating to sn_dbc::DbcContent
```
dan-da, line 8:
```suggestion
///! Unblinded data for creating sn_dbc::DbcContent
```

src/builder.rs
```diff
@@ -0,0 +1,81 @@
...
+        self.outputs.push(output);
+        self
+    }
+
+    pub fn add_outputs(mut self, outputs: impl IntoIterator<Item = Output>) -> Self {
+        self.outputs.extend(outputs);
+        self
+    }
+
+    pub fn build(self) -> Result<(ReissueTransaction, HashMap<crate::Hash, blsttc::PublicKey>)> {
```
dan-da, line 41: this is nice.  much cleaner for the caller.  I didn't like that client code had to calc and pass blinding_factor into DbcContent.

src/mint.rs
```diff
@@ -568,91 +528,78 @@ mod tests {
...
-                input_hashes.clone(),
-                1000,
-                0,
-                outputs_owner.public_key_set.public_key(),
-                genesis_secrets.blinding_factor,
-            )
-            .unwrap()]),
-        };
+        let (double_spend_reissue_tx, _output_owners) = crate::TransactionBuilder::default()
+            .add_input(genesis_dbc, genesis_secret_amounts)
```
dan-da, line 578: They are secrets (plural) related to the amount (singular).

```suggestion
            .add_input(genesis_dbc, genesis_amount_secrets)
```

src/mint.rs
```diff
@@ -796,175 +728,148 @@ mod tests {
...
-                        content,
-                        transaction: transaction.clone(),
-                        transaction_sigs: BTreeMap::from_iter(
-                            transaction_sigs
-                                .iter()
-                                .map(|(input, _)| (*input, (genesis_key, mint_sig.clone()))),
-                        ),
-                    }
-                }));
+                let output_dbcs =
```
dan-da, line 858: Would it make sense to have a DBC builder as well?

src/mint.rs
```diff
@@ -1255,22 +1168,15 @@ mod tests {
         //  b) with true committed amount (out of band for recipient).  (should succeeed)
```
dan-da, line 1168:
```suggestion
        //  b) with true committed amount (out of band for recipient).  (should succeed)
```

dan-da:
Looks good overall.    

Another rough edge I've noticed in the past is that the client must verify output of Mint::reissue() and manually build Dbc(s) from it.  I wonder if it wouldn't make sense to have a DbcBuilder as well.  I think that from client's perspective, it's nice to just call reissue() and get fully formed Dbc(s) as the output, or an error.   Or as close to that as possible.  I'm not suggesting that for this PR, just kinda thinking out loud...

oetyng:
> Looks good overall.
> 
> Another rough edge I've noticed in the past is that the client must verify output of Mint::reissue() and manually build Dbc(s) from it. I wonder if it wouldn't make sense to have a DbcBuilder as well. I think that from client's perspective, it's nice to just call reissue() and get fully formed Dbc(s) as the output, or an error. Or as close to that as possible. I'm not suggesting that for this PR, just kinda thinking out loud...

Yeah, makes sense to me.

src/builder.rs
oetyng: Very nice.

src/mint.rs
davidrusu: Yes, might be nice to have reissue return Dbc shares as we discussed in meeting.

dan-da:
lgtm, except fmt/clippy check is failing now.

dan-da (approved):
lgtm!

PR #76, opened 2021-08-18
Updates mint-repl to use new TransactionBuilder

dan-da:
Also adds some helper fns to TransactionBuilder:

* inputs_hashes()
* inputs_amount_sum()
* outputs_amount_sum()

These are useful for mint-repl, so probably would be useful for other clients.


davidrusu (approved):
Looks good

PR #80, opened 2021-08-18
creates type alias Amount to use in place of u64

dan-da:
It's been bugging me that we use u64 everywhere for amounts.   This alias would make it easier to change in the future if we ever should wish to.  (with the caveat that we are presently limited to u64 by bulletproofs::RangeProof)


davidrusu (approved):
🚢

PR #83, opened 2021-08-20
add ReissueShare and DbcBuilder to simplify aggregating reissue results from mint nodes

dan-da:
Changes:
* Mint::reissue() now returns a ReissueShare instead of a tuple
* adds DbcBuilder to aggregate ReissueShare, validate, combine signatures, and gen output DBCs
* AmountSecrets::to_bytes() takes self instead of &self.  recommended by clippy.
* updates tests, mint-repl, and reissue bench to use DbcBuilder and/or ReissueShare

----
**Discussion for a possible followup PR:**

One observation I made while implementing this is that we return PublicKeySet for every output DBC in every reissue().  Yet to combine signatures one must use a single PublicKeySet.   This means the client should/must verify that all returned PublicKeySet are the same which is a bit convoluted with this structure.  It would be a bit simpler to just return a single PublicKeySet from Mint::reissue() in the ReissueShare and omit it from MintNodeSignatures, which then becomes **BTreeMap<DbcContentHash, NodeSignature>** instead of **BTreeMap<DbcContentHash, (PublicKeySet, NodeSignature)>**.

Is there ever a situation where we would actually want distinct PublicKeySet per output Dbc per MintNode?   If so, DbcBuilder::build() would fail miserably when it encounters that...

### DIFF
--- a/src/builder.rs
+++ b/src/builder.rs
@@ -135,12 +135,12 @@ impl DbcBuilder {
     /// has not been set or no ReissueShare has been added.
     pub fn build(self) -> Result<Vec<Dbc>> {
         if self.reissue_shares.is_empty() {
-            return Ok(vec![]);
+            return Err(Error::NoReissueShares);
         }
 
         let reissue_transaction = match self.reissue_transaction {
             Some(rt) => rt,
-            None => return Ok(vec![]),
+            None => return Err(Error::NoReissueTransaction),
         };
 
         let mut mint_sig_shares: Vec<NodeSignature> = Default::default();
--- a/src/builder.rs
+++ b/src/builder.rs
@@ -1,9 +1,13 @@
+use blsttc::{PublicKeySet, SignatureShare};
 use std::collections::{BTreeSet, HashMap, HashSet};
 use std::iter::FromIterator;
 
 use curve25519_dalek_ng::scalar::Scalar;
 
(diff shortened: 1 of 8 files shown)


src/builder.rs
```diff
@@ -94,3 +98,135 @@ impl TransactionBuilder {
...
+        self
+    }
+
+    /// Build the output DBCs
+    ///
+    /// Note that the result Vec may be empty if the ReissueTransaction
+    /// has not been set or no ReissueShare has been added.
+    pub fn build(self) -> Result<Vec<Dbc>> {
+        if self.reissue_shares.is_empty() {
+            return Ok(vec![]);
```
davidrusu, line 138: I think this should be an error, the reissue transaction will likely specify output DBC's so it seems odd to me that an Ok result would not match the number of output DBC's

src/builder.rs
```diff
@@ -94,3 +98,135 @@ impl TransactionBuilder {
...
+        }
+
+        let reissue_transaction = match self.reissue_transaction {
+            Some(rt) => rt,
+            None => return Ok(vec![]),
```
davidrusu, line 143: Same here, I think this should also be an error e.g. MissingReissueTransaction

davidrusu (approved):
LGTM

PR #85, opened 2021-09-08
feat: add ReissueRequestBuilder to simplify aggregating dbc ownership proofs

dan-da:
So now creating a ReissueRequest can look something like:

```
        let rr = ReissueRequestBuilder::new(reissue_tx);
            .add_dbc_signer(
                genesis_dbc.name(),
                genesis_owner.public_key_set,
                genesis_owner.index,
                genesis_owner.secret_key_share,
            )
            .build()?;
```

or for a more complicated case with multiple input dbcs and 2 signers for each:

```
let owner_shares = vec![(0, share0),(1, share1)];
let mut builder = ReissueRequestBuilder::new(reissue_tx.clone());
for dbc in reissue_tx.inputs.iter() {
    builder.add_dbc_signers(
        dbc.name(),
        owner.public_key_set,
        owner_shares
    );
}
let reissue_request = builder.build()?;
```

### DIFF
--- a/src/builder.rs
+++ b/src/builder.rs
@@ -1,12 +1,12 @@
-use blsttc::{PublicKeySet, SignatureShare};
-use std::collections::{BTreeSet, HashMap, HashSet};
+use blsttc::{Fr, IntoFr, PublicKey, PublicKeySet, SecretKeyShare, Signature, SignatureShare};
+use std::collections::{BTreeMap, BTreeSet, HashMap, HashSet};
 use std::iter::FromIterator;
 
 use curve25519_dalek_ng::scalar::Scalar;
 
 use crate::{
-    Amount, AmountSecrets, Dbc, DbcContent, Error, Hash, NodeSignature, ReissueShare,
-    ReissueTransaction, Result,
+    Amount, AmountSecrets, Dbc, DbcContent, DbcContentHash, Error, Hash, NodeSignature,
+    ReissueRequest, ReissueShare, ReissueTransaction, Result,
 };
 
 ///! Unblinded data for creating sn_dbc::DbcContent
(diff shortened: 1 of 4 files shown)


src/builder.rs
```diff
@@ -99,6 +99,107 @@ impl TransactionBuilder {
     }
 }
 
+/// Builds a ReissueRequest from a ReissueTransaction and
+/// any number of (input) DBC hashes with associated ownership share(s).
+#[derive(Default)]
+pub struct ReissueRequestBuilder {
+    pub reissue_transaction: Option<ReissueTransaction>,
```
davidrusu, line 106: Lets make this field not Optional. Our constructor already expects a reissue transaction and it'll save us from an additional error to check

src/builder.rs
```diff
@@ -99,6 +99,107 @@ impl TransactionBuilder {
...
+        self
+    }
+
+    /// Add a single signer share for a DBC hash
+    pub fn add_dbc_signer<FR: IntoFr>(
+        mut self,
+        dbc: DbcContentHash,
+        public_key_set: PublicKeySet,
+        share_index: FR,
+        secret_key_share: SecretKeyShare,
```
davidrusu, line 132: The pattern we're following elsewhere (in the KeyManager) in the crate is that the share index and secret key share are in a tuple together when we pass it around:

```suggestion
        secret_key_share: (FR, SecretKeyShare),
```

src/builder.rs
```diff
@@ -99,6 +99,107 @@ impl TransactionBuilder {
...
+        (*entry).push((public_key_set, (share_index.into_fr(), secret_key_share)));
+        self
+    }
+
+    /// Add a list of signer shares for a DBC hash
+    pub fn add_dbc_signers<FR: IntoFr>(
+        mut self,
+        dbc: DbcContentHash,
+        public_key_set: PublicKeySet,
+        secret_key_shares: Vec<(FR, SecretKeyShare)>,
```
davidrusu, line 144: I see you follow it here as well, would be good to use the same secret share tuple pattern above.

src/builder.rs
```diff
@@ -99,6 +99,107 @@ impl TransactionBuilder {
...
+
+        for (dbc, signers) in self.signers_by_dbc.iter() {
+            let pks_set: HashSet<PublicKeySet> = signers.iter().map(|s| s.0.clone()).collect();
+            if pks_set.len() != 1 {
+                return Err(Error::ReissueRequestPublicKeySetMismatch);
+            }
+            let owner_public_key_set = match pks_set.iter().next() {
+                Some(pks) => pks,
+                None => return Err(Error::ReissueRequestPublicKeySetMismatch),
+            };
```
davidrusu, line 174: Could we modify the `signers_by_dbc` to remove these error checks? I.e. what about this:

```patch
-     pub signers_by_dbc: HashMap<DbcContentHash, Vec<(PublicKeySet, (Fr, SecretKeyShare))>>,
+     pub signers_by_dbc: HashMap<DbcContentHash, (PublicKeySet, (Fr, SecretKeyShare))>,
```

davidrusu (commented):
Looks really nice, left a few comments.

Do you plan on redoing more of the tests/examples/benches using this builder?

dan-da:
cleanup:  closing out-dated PRs.

PR #90, opened 2021-09-15
ReissueRequestBuilder

davidrusu:


src/builder.rs
```diff
@@ -97,9 +97,9 @@ impl TransactionBuilder {
 
 /// Builds a ReissueRequest from a ReissueTransaction and
 /// any number of (input) DBC hashes with associated ownership share(s).
-#[derive(Debug, Default)]
+#[derive(Debug)]
 pub struct ReissueRequestBuilder {
-    pub reissue_transaction: Option<ReissueTransaction>,
+    pub reissue_transaction: ReissueTransaction,
```
dan-da, line 102: Are you sure about this?

There are times when you might be building up a ReissueRequestBuilder and a TransactionBuilder at the same time.    For example, mint-repl does this when collecting user input.  (in blind sigs branch)

davidrusu: I've gone through all use-cases we have in the repo (mint-repl / benches / tests) and we always have access to the transaction at the time that you want to start using the `ReissueRequestBuilder`.

Happy to go back once we find a case where it falls apart, as it is we get to remove one error check/variant

dan-da: yeah, I didn't really like the Option myself, but it serves a purpose.  Did you see the mint-repl link I posted above?  anyway, here's the code:

```
    let mut tx_builder: TransactionBuilder = Default::default();
    let mut rr_builder = ReissueRequestBuilder::default();
    let mut pk_pks: HashMap<PublicKey, PublicKeySet> = Default::default();

    // Get from user: input DBC(s) and required # of SecretKeyShare+index for each.
    loop {
        println!("--------------");
        println!("Input DBC #{}", tx_builder.inputs.len());
        println!("--------------\n");

        let dbc_input = readline_prompt_nl("\nDBC Data, or 'done': ")?;
        let dbc: DbcUnblinded = if dbc_input == "done" {
            break;
        } else {
            from_be_hex(&dbc_input)?
        };

        println!(
            "We need {} SecretKeyShare in order to spend the input.",
            dbc.owner.threshold() + 1
        );

        while rr_builder.num_signers_by_dbc(dbc.inner.name()) < dbc.owner.threshold() + 1 {
            let key = readline_prompt_nl("\nSecretKeyShare, or 'cancel': ")?;
            let secret: SecretKeyShare = if key == "cancel" {
                println!("\nreissue_ez cancelled\n");
                return Ok(());
            } else {
                from_be_hex(&key)?
            };
            let idx_input = readline_prompt("\nSecretKeyShare Index: ")?;
            let idx: usize = idx_input.parse()?;

            rr_builder =
                rr_builder.add_dbc_signer(dbc.inner.name(), dbc.owner.clone(), idx, secret.clone());
        }

        tx_builder = tx_builder.add_input(dbc.inner.clone());
    }

// ... a bunch of stuff ...

    let (transaction, output_secrets) = tx_builder.build()?;

// .. more stuff ...

    rr_builder = rr_builder.set_reissue_transaction(transaction);
    let reissue_request = rr_builder.build()?;
```

So with your change, it would be necessary to collect all the dbc signing stuff in some separate data structure, then pass that into the builder.   This usage seems cleaner and more natural to me.

I guess if you can re-work the mint-repl code above in a natural way that doesn't require a separate temp collection I'd be more convinced of the change, but I don't immediately see how to do that.

I feel like real users would have other/similar use cases for it...

davidrusu: Ah, I didn't port that bit of code from the blinded sig branch, ok I'll port that over now and see how it goes.

davidrusu: @dan-da it worked out, we can delay creating the ReissueRequestBuilder until we have the transaction: https://github.com/maidsafe/sn_dbc/pull/90/commits/f29525a4e7f8d034e748eb237e9344178bb5e508

dan-da: yeah, the inputs_sks is the temp collection that I eliminated by using the ReissueRequestBuilder.    So that's cheating.  :D

> let mut inputs_sks: HashMap<DbcUnblinded, BTreeMap<usize, SecretKeyShare>> = Default::default();

davidrusu: How about we change when we ask for secrets:

1. ask for inputs and outputs
2. build transaction
3. ask for secrets for each input in order to sign the transaction
4. build the reissue request
5. send the request
6. build the output dbcs

dan-da: not natural. 

The mint-repl is written the way it is because the user is in the mental state of dealing with the first input, then the 2nd, and so on.

What you are suggesting is to change the use case.   But I don't think that's right.   The use case is the use case.

I built the ReissueRequestBuilder with the use case in mind, to simplify it.    I think it is representative of what other developers will want/need to do, outside of test cases.

The builder as I wrote it is very flexible, because one can just instantiate it and then add things as they are available.   That actually follows the pattern of the first builder you made.  It was a conscious decision, thinking about these issues, not a mistake.  The point is to simplify the building task for api user, not complicate them and force jumping through unnecessary hoops.

davidrusu: This ^^ is a bit more involved. I don't think I want to make that change right now.

This flow will change again once we go to "client-writes-spendbook".

Personally, I'm happy to trade off a temporary map in reissue_ez for no Option and no error checks/variant in the builder.

I think we'll be able to get rid of the temporary variable here once we go to client-writes-spendbook since we only need the secrets to do the spendbook write, we don't need them for the reissue request.

davidrusu: Alright, I'll add back the Option

dan-da: hmm, well for master, yeah do as you wish.   For blind sigs branch I want to keep it as I wrote it until such time as there is no longer a need to collect keys for the ReissueRequest.    (at which point, do we even need a builder?  It's basically just the ReissueTransaction right?

dan-da: btw, a related but orthogonal thing I've thought about a few times is that we are combining both client and server generated errors into one Error struct.  we probably should split it up, eg MintNodeError and DbcClientError, or something like that.

dan-da (dismissed):
lgtm!    edit:  except I see there is a cargo/fmt test failing.

davidrusu:
> edit: except I see there is a cargo/fmt test failing.
There was an odd caching issue on github actions, forced a cache eviction and it passed

dan-da (approved):
ship it!

PR #93, opened 2021-09-16
GenesisShares

davidrusu:
This factors out the genesis shares work from the Blinded Sigs work so that we can get it on master earlier and look at the blinded sigs work on it's own.


dan-da (approved):
ship it!

PR #96, opened 2021-09-21
Client writes spentbook

davidrusu:
This PR modifies the reissue flow to have client write to the spentbook prior to asking mints to sign the reissue.

This makes the Mint's task read-only, the Mint only checks the spentbook for transaction commitments for each input DBC to verify that all inputs are committed to the same transaction and validates the transaction balances.

Some notes
- For mint-repl, the `sign_tx` step is still implemented, but it isn't used, perhaps this should be a `log_tx_in_spentbook` command instead?
- `ReissueRequestBuilder` has been removed, mostly because we don't need to collect ownership proofs anymore. This task is delegated to the Spentbook implementation. We still need something to replace it as we will need to aggregate signatures for a proper Spentbook, I've sidestepped this for now in our toy example by not requiring signatures to write to our in-memory Spentbook.


src/builder.rs
```diff
@@ -95,123 +95,6 @@ impl TransactionBuilder {
     }
 }
 
-/// Builds a ReissueRequest from a ReissueTransaction and
-/// any number of (input) DBC hashes with associated ownership share(s).
-#[derive(Debug, Default)]
-pub struct ReissueRequestBuilder {
```
dan-da, line 101: It seems to me that the client will still need a similar builder for aggregating signature shares and generating the spendbook entry, no?

davidrusu: yes.. maybe that should be part of blsttc? a more general purpose Signature Share Aggregator?

dirvine: That would be nice indeed @iancoleman ;-)

src/error.rs
```diff
@@ -88,14 +78,15 @@ pub enum Error {
     #[error("Invalid Amount Commitment")]
     AmountCommitmentInvalid,
 
+    #[error("Double spend detected")]
```
dan-da, line 81: hmm, I thought in our conversation this morning you said that Double Spend prevention is now a spentbook responsibility.  So I'm surprised to see this added...?

src/error.rs
```diff
@@ -33,21 +30,12 @@ pub enum Error {
...
     MissingSignatureForInput,
-    #[error("At least one input is missing an ownership proof")]
-    MissingInputOwnerProof,
+    #[error("At least one input is missing a spent proof for {0:?}")]
+    MissingSpentProof(SpendKey),
+    #[error("Invalid SpentProof Signature for {0:?}")]
+    InvalidSpentProofSignature(SpendKey),
     #[error("Mint request doesn't balance out sum(input) == sum(output)")]
     DbcReissueRequestDoesNotBalance,
-    #[error("Failed to unblind an input DBC")]
```
dan-da, line 40: nice to see error cases being removed.  :-)

src/mint.rs
```diff
@@ -131,8 +131,7 @@ impl ReissueTransaction {
 #[derive(Eq, PartialEq, Debug, Clone, Deserialize, Serialize)]
 pub struct ReissueRequest {
     pub transaction: ReissueTransaction,
-    // Signatures from the owners of each input, signing `self.transaction.blinded().hash()`
-    pub input_ownership_proofs: HashMap<SpendKey, blsttc::Signature>,
+    pub spent_proofs: BTreeMap<SpendKey, SpentProof>,
```
dan-da, line 134: We are mixing present and past tense here:   SpendKey and SpentProof.

Perhaps SpendKey should be renamed to SpentKey?     It is a little bit weird though since a SpentKey is first generated by client before dbc is actually written to spentbook (spent).

I'm not sure what the right answer is.  just raising for consideration.

src/mint.rs
```diff
@@ -142,21 +141,16 @@ pub struct ReissueShare {
 }
 
 #[derive(Debug, Clone, Deserialize, Serialize)]
-pub struct Mint<K, S>
+pub struct MintNode<K>
```
dan-da, line 144: Yay!

src/mint.rs
```diff
@@ -293,16 +230,6 @@ impl<K: KeyManager, S: SpendBook> Mint<K, S> {
             ))),
         ))
     }
-
```
dan-da, line 296: Yay!

src/mint.rs
```diff
@@ -440,79 +381,6 @@ mod tests {
         Ok(())
     }
 
-    #[test]
```
dan-da, line 443: Scary to lose this test.

src/mint.rs
```diff
@@ -521,10 +389,10 @@ mod tests {
         output_amounts: TinyVec<TinyInt>,
         // Controls which output dbc's will receive extra parent hashes
         extra_output_parents: TinyVec<TinyInt>,
-        // Include a valid ownership proof for the following inputs
-        input_owner_proofs: TinyVec<TinyInt>,
-        // Include an invalid ownership proof for the following inputs
-        invalid_input_owner_proofs: TinyVec<TinyInt>,
+        // Include a valid SpendProof for the following inputs
```
dan-da, line 392: SpentProof

src/mint.rs
```diff
@@ -521,10 +389,10 @@ mod tests {
...
+        valid_spent_proofs: TinyVec<TinyInt>,
+        // Include an invalid SpendProofs for the following inputs
```
dan-da, line 394: SpentProof.

src/mint.rs
```diff
@@ -591,28 +458,43 @@ mod tests {
             )
             .build()?;
 
+        let spend_sig = genesis_owner.public_key_set.combine_signatures(vec![(
```
dan-da, line 461: spent_sig

src/mint.rs
```diff
@@ -1030,24 +925,38 @@ mod tests {
...
                 owner: outputs_owner_pk,
             })
             .build()?;
 
-        let rr = ReissueRequestBuilder::new(transaction)
-            .add_dbc_signer(
-                input_dbc.spend_key(),
-                outputs_owner.public_key_set.clone(),
-                (outputs_owner.index, outputs_owner.secret_key_share.clone()),
+        let spend_sig = genesis_owner.public_key_set.combine_signatures(vec![(
```
dan-da, line 936: spent_sig.     maybe search/replace.  Spend/Spent,  spend/spent.

src/spent_proof.rs
```diff
@@ -0,0 +1,106 @@
...
+        )
+    }
+}
+
+/// A share of a SpentProof, combine enough of these to form a
+/// SpentProof.
+#[derive(Debug, Clone, PartialEq, Eq, Hash, Serialize, Deserialize)]
+pub struct SpentProofShare {
+    /// Signature from dbc.spend_key() over the transaction
+    pub spend_sig: Signature,
```
dan-da, line 48: spent_sig

dan-da (dismissed):
Looks good overall.  There seems a tension between tense of spendkey and spentbook.  Maybe we should just standardize on spent to avoid confusion.

src/mint.rs
davidrusu: yep, not great.

How about `DBC SpendingKey` in-place of `DBC SpendKey` and Spent for everything else: `SpentSig`/`SpentBook`/`SpentProof`?

dan-da: That could be...  Seems about the same meaning as SpendKey though.

is it terrible to just call it SpentKey, so we are past tense with everything?

In a sense it *is* a spent key, as its basically meaningless unless/until actually stored in the SpentBook.

but I'm ok whichever way you decide to go.

davidrusu: I'd rather go the other way, make them all use the Spend* prefix

    SpendKey, SpendBook, SpendProof, SpendSig

but.. I don't want to make this change in this PR. Lets keep it as is for now

davidrusu:
@dan-da ready for another look.

src/mint.rs
dan-da: > make them all use the Spend* prefix

yeah, I'd be good with that.

dan-da (approved):
lgtm!

PR #97, opened 2021-09-21
feat: simplify Denomination to use powers-of-ten with exponent

dan-da:
See the lengthy comment in amount.rs for explanation of the approach.  The biggest thing is that Amount changes from a u128 alias to struct Amount{ count: u32, unit: i8 }.  This is a different way of representing transactable amounts.

This is a draft PR because mint-repl and bench are not working yet.   Also the approach is still being discussed by the team.


src/amount.rs
```diff
@@ -0,0 +1,687 @@
...
+        Self { count, unit }
+    }
+
+    // note: It's recommended to make this larger than Mint's genesis amount,
+    //       else one immediately gets AmountIncompatible errors when reissuing
+    //       two outputs: [1, GenesisAmount - 1] and must instead reissue to
+    //       larger denoms/units.  not a big deal, but can be confusing when
+    //       writing test cases.
+    pub fn counter_max() -> AmountCounter {
+        // One billion units per tx ought to be enough for anybody! -- danda 2021.
```
davidrusu, line 120: 😆

src/amount.rs
```diff
@@ -0,0 +1,687 @@
...
+/// by reissuing in ever smaller amounts, but wallet software should generally
+/// be trying NOT to do that.  And if a user does manage it, then s/he will have
+/// difficulty using them in transactions with other people.  Fortunately the
+/// reverse process can be used to bring them up into a "normal" range again.
+///
+#[derive(Clone, Debug, Copy, Default, Serialize, Deserialize)]
+pub struct Amount {
+    pub count: AmountCounter,
+    pub unit: PowerOfTen,
+}
```
davidrusu, line 92: the range of values this `Amount` struct can take is `domain(AmountCounter) * domain(PowerOfTen)` which gets very large if the range of `AmountCounter` is 1billion, with a single power of ten we have 1billion anonymity classes right? That would seem like enough to break traceability.

Can we restrict this count down closer to 0..9 to get 10 anonymity classes per magnitude?

dan-da: yeah, so the Amount struct is representing transactable amounts, such as prices on an item.   Using USD as an example, if you sell me your old laptop and you ask $357 for it, well there is no $357 bill.  So instead I give you three `100` one `50` one `5` and two `one`.    Taking this back to our DBC denom scheme, I would give you one `300`, one `50` and one `7`.  There is always just one coin/bill per digit.

So the anonymity pools being used are those of the denominations:  a pool of `300`, another of `50` and another of `7`.   The pool for each is the set of all transactions that use the same denomination.

Now one can say that the transacted amount itself, 357 is a unique and interesting number.   And it is.  However it is unique to that single transaction.  So an observer can see that $357 was transacted there, but cannot link to any other tx by 357.  Only by the 300, 50, and 7, which are in very common (constant) usage by others.

dan-da: We can certainly lower the counter_max() number.  Doing so would reduce outliers, eg the number of people transacting 1 billion USD is going to be pretty small, so those tx could stand out.  However, it also forces people to use less granular amounts in their transactions.  This might annoy people, so I don't think we should get too carried away there.  Maybe we could consider going as low as 10^6.   Anyway, it's a knob we can twist.... lowering it both increases privacy (few outliers) and reduces work for the mint.

PR #99, opened 2021-10-02
Amount fixes and tests pr

dan-da:
improves correctness and comments for Amount, Denomination and adds test cases for Amount.

### DIFF
--- a/src/amount.rs
+++ b/src/amount.rs
@@ -1,8 +1,4 @@
 use crate::{Error, Result};
-use rug::ops::DivRounding;
-use rug::ops::Pow;
-use rug::Integer;
-use rug::Rational;
 use serde::{Deserialize, Serialize};
 use std::cmp::Ordering;
 use std::collections::BTreeMap;
@@ -92,9 +88,9 @@ pub struct Amount {
     unit: PowerOfTen,
 }
 
-/// A NormalizedAmount is just like an Amount except that count is a Big Integer.
-/// So sum or difference of any two Amounts sharing the same unit can be represented
-/// with a NormalizedAmount.
+/// A NormalizedAmount is just like an Amount except that count is a u128
+/// So sum or difference of any two compatible Amounts sharing the same unit can
+/// be represented with a NormalizedAmount.
 ///
 /// For now, have this only for internal use/ops.
 #[derive(Debug, Default)]
@@ -173,15 +169,6 @@ impl Amount {
         // represented by denominations.
         debug_assert!(10u32.pow(Self::COUNT_MAX_TEN_POW as u32) == Self::COUNT_MAX);
         PowerOfTen::MIN + Self::COUNT_MAX_TEN_POW
-    }
-
-    /// Converts Amount to a rug::Rational number.
-    //
-    // todo: we should think about if we want to expose rug::Rational
-    //       in our public API or not. Possibly this could be exposed
-    //       in a companion helper crate.
-    pub fn to_rational(self) -> Rational {
-        Rational::from(10).pow(self.unit as i32) * Rational::from(self.count)
     }
 
     /// generates a mapping of PowerOfTen to SI names.
@@ -308,7 +295,7 @@ impl Amount {
         let mut unit = self.unit;
         while count % 10 == 0 && unit < Self::unit_max() {
             unit += 1;
-            count = count.div_ceil(10);
+            count /= 10;
         }
 
         debug_assert!(count <= Self::counter_max());
@@ -328,7 +315,7 @@ impl Amount {
     // count = 255,  unit = 1    = 2550.
     //
     // Because count can overflow in one of the Amount, we return
-    // NormalizedAmount that uses a big rug::Integer for the count.
+    // NormalizedAmount that uses a u128 for the count.
     fn normalize(a: Self, b: Self) -> Result<(NormalizedAmount, NormalizedAmount)> {
         let a = a.to_highest_unit();
         let b = b.to_highest_unit();
@@ -390,7 +377,7 @@ impl Amount {
     /// performs addition operation and returns error if operands are incompatible.
     pub fn checked_add(self, other: Self) -> Result<Self> {
         // steps:
-        // 1. normalize to same units.  use rug:Integer to represent count.
+        // 1. normalize to same units.  use u128 to represent count.
         // 2. add counts.
         // 3. find unit in which count is less than Self::counter_max()
         // 4. Amount::new()
@@ -424,7 +411,7 @@ impl Amount {
         }
 
         // steps:
-        // 1. normalize to same units.  use rug:Integer to represent count.
+        // 1. normalize to same units.  use u128 to represent count.
         // 2. subtract count.
         // 3. find unit in which count is less than Self::counter_max()
         // 4. Amount::new()
@@ -462,14 +449,6 @@ impl Amount {
     // attempts to parse u128 into Amount
     fn from_str_u128(s: &str) -> Result<Self> {
         match s.parse::<u128>() {
-            Ok(v) => Self::try_from(v),
-            Err(_) => Err(Error::AmountUnparseable),
-        }
-    }
-
-    // attempts to parse rug::Rational into Amount
-    fn from_str_rational(s: &str) -> Result<Self> {
-        match s.parse::<Rational>() {
             Ok(v) => Self::try_from(v),
             Err(_) => Err(Error::AmountUnparseable),
         }
@@ -536,17 +515,8 @@ impl FromStr for Amount {
     type Err = Error;
 
     fn from_str(s: &str) -> Result<Self, Self::Err> {
-        // note: from_str_u128 is an optimization to avoid calling
-        //       presumed slower from_str_rational() if not necessary.
-        //       perhaps the optimization is not worth the extra
-        //       code/complexity.
+        // parses eg 2345322200000000
         if let Ok(v) = Self::from_str_u128(s) {
-            return Ok(v);
-        }
-
-        // parses eg 0, 1, 10, 1/10, 1/5, 2/5
-        //           15000000000000000000000000000000000000000000000000000000000000000
-        if let Ok(v) = Self::from_str_rational(s) {
             return Ok(v);
         }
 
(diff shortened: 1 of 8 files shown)


src/amount.rs
```diff
@@ -266,10 +357,8 @@ impl Amount {
             } else {
                 (b.unit..a.unit).len() as u32
             };
-            let unit_base = *[a.unit, b.unit].iter().min().unwrap();
-
-            // println!("unit_distance_range: {:?}", (a.unit..b.unit));
-            // println!("unit_distance: {}", unit_distance);
+            // note: this unwrap_or() can never fail because there are two items.
+            let unit_base = *[a.unit, b.unit].iter().min().unwrap_or(&0);
```
davidrusu, line 361: You can avoid all of the unwraps in this section by using [`std::cmp::min_by`](https://doc.rust-lang.org/std/cmp/fn.min_by.html):

```rust
let (base, other) = std::cmp::min_by(a, b, |a, b| a.unit.cmp(&b.unit));
let norm_base = NormalizedAmount {
    count: Integer::from(base.count),
    unit: base.unit
};

let norm_other = NormalizedAmount {
    count: Integer::from(10).pow(other.unit - base.unit),
    unit: base.unit
 }

(norm_base, norm_other)

File: src/amount.rs [hunk 2]
```diff
@@ -208,36 +292,43 @@ impl Amount {
...
     // count = 255, unit = 1    = 2550.
-
-    // option a:
+    //
+    // if we normalize to highest unit:
     // count = 25, unit = 2    = 25 * 100 = 2500
-    // count = 25, unit = 2    = 25 * 10 = 2500    <---- loses information.
-
-    // option b:
+    // count = 25, unit = 2    = 25 * 10 = 2500    <---- loses information. can't do this.
```
davidrusu, line 324:
```suggestion
    // count = 25, unit = 2    = 25 * 100 = 2500    <---- loses information. can't do this.
```

src/amount.rs
```diff
@@ -168,19 +211,46 @@ impl Amount {
         ]
         .iter()
         .cloned()
```
davidrusu, line 213: Can you use `into_iter` here?
```suggestion
        .into_iter()
```

src/amount.rs
```diff
@@ -87,63 +88,105 @@ pub type AmountCounter = u32;
...
+        // represented by denominations.
+        debug_assert!(10u32.pow(Self::COUNT_MAX_TEN_POW as u32) == Self::COUNT_MAX);
+        PowerOfTen::MAX - Self::COUNT_MAX_TEN_POW
     }
 
-    pub fn unit_min() -> i8 {
-        -Self::unit_max()
+    /// returns minimum possible value for unit field
+    pub fn unit_min() -> PowerOfTen {
+        // We increase max unit by size of counter max so that
```
davidrusu, line 169:
```suggestion
        // We increase min unit by size of counter max so that
```

src/amount.rs
```diff
@@ -87,63 +88,105 @@ pub type AmountCounter = u32;
...
     unit: PowerOfTen,
 }
 
 impl Amount {
-    pub fn new(count: AmountCounter, unit: PowerOfTen) -> Self {
-        // We constrain count to ::counter_max().  If you want to use a bigger value,
-        // you must change the unit.
-        debug_assert!(count <= Self::counter_max());
+    /// maximum value for Self::count. MUST equal 10^COUNT_MAX_TEN_POW
+    //  note: we define these separately to avoid unnecessary runtime 10.pow() call.
```
davidrusu, line 108: [`pow`](https://doc.rust-lang.org/nightly/std/primitive.i64.html#method.pow) is a const function so 10.pow() willl be evaluated at compile time, it shouldn't impact runtime

dan-da:
@davidrusu thx for the comments.  just fyi, I will be updating this with a substantial simplification soon, and will incorporate your suggestions. 

I have further refactored and was able to remove rug dependency entirely.... ie no more rug::Rational, rug::Integer.   I like it because it reduces the code size/footprint in sn_dbc, all integer math and no need to compile gmp library.

The main things we lose are:
1. the ability to parse fractions and repr/display fractions.
2. Amount::to_rational(), which was quite useful as a double check in test cases, so I'm going to add test vectors for checked_add and checked_subtract instead.

But these are not things the Mint needs and perhaps not most clients.

Anyway, I'm creating a helper crate for client use (possibly) that has the removed logic in an AmountRational struct.  Just to keep the code around in case it's useful at some point.... it took me a lot of work and head scratching to get `impl From<Rational> for Amount` working.  ;)

davidrusu:
Thanks for the heads up, let me know when the new work is ready for review

src/amount.rs
dan-da: yep!

src/amount.rs
dan-da: min_by is nice, but `let (base, other)` seems like black magic to me since min_by returns a single value T of same type as the input params.   anyway, this code has changed a bit and I came up with a condensed version.

src/amount.rs
dan-da: whoops, I mean no!    I tried that before also, and get:

```
error[E0277]: a value of type `BTreeMap<amount::Amount, &str>` cannot be built from an iterator over elements of type `&(amount::Amount, &str)`
   --> src/amount.rs:983:10
    |
983 |         .collect()
    |          ^^^^^^^ value of type `BTreeMap<amount::Amount, &str>` cannot be built from `std::iter::Iterator<Item=&(amount::Amount, &str)>`
    |
    = help: the trait `std::iter::FromIterator<&(amount::Amount, &str)>` is not implemented for `BTreeMap<amount::Amount, &str>`
```

src/amount.rs
davidrusu: hehe, yes, idk what I was thinking

dan-da:
@davidrusu ready for review when you have a chance.

src/amount.rs
dan-da: doing that now.

src/amount.rs
```diff
@@ -87,63 +88,100 @@ pub type AmountCounter = u32;
...
 
+    // only for use in this crate because code in this impl depends on
+    // count always being <= Self::counter_max().
+    pub(crate) fn new_unchecked(count: AmountCounter, unit: PowerOfTen) -> Self {
+        debug_assert!(count <= Self::counter_max());
         Self { count, unit }
     }
 
+    // returns count of 10^unit
+    #[inline]
```
davidrusu, line 134: You've added `#[inline]` to a lot of these functions, I'm curious if you noticed an impact on performance?

src/amount.rs
```diff
@@ -166,21 +204,48 @@ impl Amount {
...
+    // note: There is no SI name for 10^0, so those are just formatted
+    //       as regular integer values via ::to_string()
+    // note: SI names only extend from 10^-24..10^24.  Outside that range
+    //       we fallback to formatting with ::to_string()
+    pub fn to_si_string(self) -> String {
+        let map = Self::si_map();
+        // note: this unwrap_or can never fail because map always contains values.
+        // note: we use max+2 to account for eg 10 yotta, 100 yotta.
+        let min = map.keys().min().unwrap_or(&0);
+        let max = map.keys().max().unwrap_or(&0) + 2;
```
davidrusu, line 225:
```suggestion
        let min_si_unit = map.keys().min().unwrap_or(&0);
        let max_si_unit = map.keys().max().unwrap_or(&0) + 2;
```

src/amount.rs
```diff
@@ -166,21 +204,48 @@ impl Amount {
...
 
-        if self.unit >= -24 && self.unit <= 24 && self.count != 0 {
+        if self.unit >= *min && self.unit <= max && self.count != 0 {
             let mut unit = self.unit;
             loop {
                 if let Some(name) = map.get(&unit) {
-                    let diff = self.unit.abs() - unit.abs();
-                    let udiff = 10u64.pow(diff as u32);
-                    let newcount = self.count as u64 * udiff;
+                    let diff = (self.unit.abs() - unit.abs()).abs();
```
davidrusu, line 235: I think you want this instead:
```suggestion
                    let diff = (self.unit - unit).abs();
```

src/amount.rs
```diff
@@ -353,48 +437,172 @@ impl Amount {
...
+impl TryFrom<u128> for Amount {
+    type Error = Error;
+
+    fn try_from(n: u128) -> Result<Self, Self::Error> {
+        let (unit, count) = calc_exponent_u128(n).ok_or(Error::AmountUnparseable)?;
+
+        let count = AmountCounter::try_from(count).map_err(|_| Error::AmountUnparseable)?;
+        Amount::new(count, unit)
+    }
+}
```
davidrusu, line 544: This is just a thought, haven't tested it, but you may be able to get all of this `TryFrom` impls for free with:

```rust
impl<T: Into<u128>> TryFrom<T> for Amount {
    type Error = Error;

    fn try_from(n: T) -> Result<Self, Self::Error> {
        let (unit, count) = calc_exponent_u128(n.into::<u128>()).ok_or(Error::AmountUnparseable)?;

        let count = AmountCounter::try_from(count).map_err(|_| Error::AmountUnparseable)?;
        Amount::new(count, unit)
    }
}
```

src/denomination.rs
```diff
@@ -1,6 +1,11 @@
+
```
davidrusu, line 1: It looks like this file is still excluded from cargo format, there's a few formatting issues in here (extra lines, etc..)

src/amount.rs
dan-da: done.

davidrusu:
Still have a few clippy failures

src/amount.rs
dan-da: ok, I guess we'll leave it as-is then.

dan-da:
> Still have a few clippy failures

hmm, do you mean in the bench maybe?   looks clean here.

```
$ cargo clippy
    Finished dev [unoptimized + debuginfo] target(s) in 0.36s
$ cargo clippy --tests
    Checking sn_dbc v2.7.1 (/home/danda/dev/maidsafe/sn_dbc)
    Finished dev [unoptimized + debuginfo] target(s) in 1.83s
```

davidrusu (approved):

PR #100, opened 2021-10-05
store mint's root public key in Dbc rather than derived-from-denom key

dan-da:
Previously, Dbc::mint_public_key contained a pk derived from the Mint's root public key using Dbc::denomination().to_bytes() as the derivation index.

This had the nice property that one could simply perform:  `self.mint_public_key.verify(self.signature)`

However, given only the derived key, one cannot determine the Mint's public key, which can be useful for validating the mint's authority.

It would be possible to simply add the root pk in addition to the derived key, however that seems wasteful because it enlarges the DBC and the derived key can be calculated.

So the approach taken here is to store the root key and derive the denomination key when signing or validating.  Because the pk and signature stored in the Dbc are no longer "paired", I renamed mint_public_key to mint_root_public_key to hopefully show this distinction and also added a comment about it.

**Performance Impact:**

This change require that the Mint node perform an extra ::derive_child() operation for each input dbc when validating.  Also clients have to derive when building, but I don't think we would care about that.

If we later wish to optimize for speed instead of size, then we could consider storing both PKs in the Dbc instead.

**Unresolved**

One thing I don't understand yet, and would like more eyes on @davidrusu @iancoleman before merging is that when the client is combining mint SignatureShares, it does not appear to matter if we use the root PublicKeySet or derived PublicKeySet.

For example, we can comment out builder.rs line 266 and/or mint.rs line 277, and the resulting DBCs still validate, all tests pass.  Does that seem correct?

### DIFF
--- a/src/builder.rs
+++ b/src/builder.rs
@@ -240,7 +240,7 @@ impl DbcBuilder {
         if pk_set.len() != 1 {
             return Err(Error::ReissueSharePublicKeySetMismatch);
         }
-        let mint_public_key_set = match pk_set.iter().next() {
+        let mint_root_public_key_set = match pk_set.iter().next() {
             Some(pks) => pks,
             None => return Err(Error::ReissueSharePublicKeySetMismatch),
         };
@@ -261,17 +261,16 @@ impl DbcBuilder {
                 );
             }
 
-            let denom_idx = dbc_envelope.denomination.to_bytes();
-            let mint_derived_pks = mint_public_key_set.derive_child(&denom_idx);
-
             // Combine signatures from all the mint nodes to obtain Mint's Signature.
-            let mint_sig = mint_derived_pks.combine_signatures(&mint_sig_shares)?;
+            let mint_signature = mint_root_public_key_set
+                .derive_child(&dbc_envelope.denomination.to_bytes())
+                .combine_signatures(&mint_sig_shares)?;
 
             // Form the final output DBCs, with Mint's Signature for each.
             let dbc = Dbc {
                 content: output_secret.dbc_content,
-                mint_public_key: mint_derived_pks.public_key(),
-                mint_signature: mint_sig,
+                mint_root_public_key: mint_root_public_key_set.public_key(),
+                mint_signature,
             };
 
             output_dbcs.push(dbc);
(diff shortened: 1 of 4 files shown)


src/builder.rs
```diff
@@ -240,7 +240,7 @@ impl DbcBuilder {
         if pk_set.len() != 1 {
             return Err(Error::ReissueSharePublicKeySetMismatch);
         }
-        let mint_public_key_set = match pk_set.iter().next() {
+        let mint_root_public_key_set = match pk_set.iter().next() {
```
davidrusu, line 243: `mint_public_key_set` was a good name, lets not introduce a new "root" key terminology.

src/dbc.rs
```diff
@@ -10,10 +10,15 @@ use crate::{DbcContent, Denomination, Error, KeyManager, PublicKey, Result, Sign
...
+/// mint_root_public_key is the root pubkey, from which
+/// denomination-specific pubkeys are derived.
+///
+/// mint_signature is a signature that can be verified with:
+/// mint_root_public_key.derive_child(&self.denomination().to_bytes()).verify(self.mint_signature)
 #[derive(Debug, Clone, PartialEq, Eq, Hash, Deserialize, Serialize)]
 pub struct Dbc {
     pub content: DbcContent,
-    pub mint_public_key: PublicKey,
+    pub mint_root_public_key: PublicKey,
```
davidrusu, line 21: Again, `mint_public_key` was fine, arguably we were putting a `mint_denomination_key` here instead of the `mint_public_key`

davidrusu:
> when the client is combining mint SignatureShares, it does not appear to matter if we use the root PublicKeySet or derived PublicKeySet.

Yep, I had the same question a week or so ago and traced through blsttc/threshold_crypto code. Turns out the crypto behind the aggregation doesn't need a public-key-set, it just needs the count of keys in the set, public key set is just a convenient place to get that key count from

The key count is the same for the derived key sets as the parent key sets, so signature aggregation still works as you'd expect.

src/builder.rs
dan-da: I changed it back.

src/dbc.rs
davidrusu: 👍 on `mint_denomination_signature`

davidrusu (approved):
🚢 !

PR #105, opened 2021-10-14
Dbc packet pr

dan-da:
Adds a DbcPacket struct and DerivedKeySet struct that together facilitate sender to create Dbc owner key from a well-known key and a random derivation index, and then to transmit the well-known pubkey, optional secret key, and derivation index to the recipient without the mint or spentbook ever seeing that data.

The DbcPacket also contains a memo field.  There are some interesting questions to resolve around that field.  See comments in the code, and we can discuss.

There is also the question of naming for DbcPacket and DerivedKeySet.  I think they are ok names, but open to suggestions...

### DIFF
--- a/src/dbc_packet.rs
+++ b/src/dbc_packet.rs
@@ -1,0 +1,211 @@
+// Copyright 2021 MaidSafe.net limited.
+//
+// This SAFE Network Software is licensed to you under The General Public License (GPL), version 3.
+// Unless required by applicable law or agreed to in writing, the SAFE Network Software distributed
+// under the GPL Licence is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, either express or implied. Please review the Licences for the specific language governing
+// permissions and limitations relating to use of the SAFE Network Software.
+
+use crate::{Dbc, Error, PublicKey, Result};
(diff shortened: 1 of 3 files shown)


dan-da:
hmm, I notice that the prop_dbc_transaction_many_to_many test failed on MacOS, but tests passed on Ubuntu.

This PR didn't change anything related.   @davidrusu you might want to take a look at that.

src/dbc_packet.rs
```diff
@@ -0,0 +1,211 @@
...
+    fn from(secret_key_set: SecretKeySet) -> Self {
+        Self {
+            public_key: secret_key_set.public_keys().public_key(),
+            derivation_index: Self::random_derivation_index(),
+            secret_key_set: Some(secret_key_set),
+        }
+    }
+}
+
+type Memo = Option<[u8; 32]>;
```
davidrusu, line 79: For a client only field, this size seems quite limiting. Can we relax this to Vec<u8>? perhaps with some (large) enforced cap?

dan-da: I went ahead and removed the field for now, per our conversation.  It can go into a future Payment struct.   I don't presently have a strong use-case example that requires a memo per DBC.

davidrusu (approved):

PR #109, opened 2021-10-21
Fix many-to-many test handling of invalid + valid proofs

davidrusu:
The many-to-many test case wasn't handling the case where both invalid and valid spentproofs were being generated.

Adapt the test case so that we generate either valid or invalid spentproofs, but not both


dan-da (approved):
lgtm.

PR #113, opened 2021-10-25
feat: make mint-repl build and work with blind sig mint

dan-da:
I tested both reissue_ez path and prepare_tx --> sign_tx --> prepare_reissue --> reissue path.


davidrusu (approved):
LGTM

PR #114, opened 2021-11-25
Amount secret pr

dan-da:
Changes to support sn_payment crate:

1. fix: adds commitment and range_proof fields to DbcContent hash().
2. Add hash fn to DerivedKeySet, DbcPacket.
3. move encrypt, decrypt, gen pedersen commitments into AmountSecret

I think it makes the API a bit cleaner and more "rusty" overall.

### DIFF
--- a/src/dbc_content.rs
+++ b/src/dbc_content.rs
@@ -19,9 +19,8 @@ use rand8::rngs::OsRng;
 use serde::{Deserialize, Serialize};
 use std::collections::BTreeMap;
 use std::convert::TryFrom;
-use tiny_keccak::{Hasher, Sha3};
 
-use crate::{Error, Hash, SpendKey};
+use crate::{ByteHash, Error, SpendKey};
 
 pub(crate) const RANGE_PROOF_BITS: usize = 64; // note: Range Proof max-bits is 64. allowed are: 8, 16, 32, 64 (only)
                                                //       This limits our amount field to 64 bits also.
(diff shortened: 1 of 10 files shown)


src/dbc_content.rs
```diff
@@ -67,6 +68,14 @@ impl AmountSecrets {
...
     }
 
+    pub fn to_pedersen_commitment(&self) -> RistrettoPoint {
+        PedersenGens::default().commit(Scalar::from(self.amount), self.blinding_factor)
+    }
+
+    pub fn to_pedersen_commitment_compressed(&self) -> CompressedRistretto {
+        self.to_pedersen_commitment().compress()
+    }
+
```
davidrusu, line 78: I think just `to_pedersen_commitment()` is enough, if someone wants the Compressed point, they can compress it themselves

src/dbc_content.rs
```diff
@@ -67,6 +68,14 @@ impl AmountSecrets {
         }
     }
 
+    pub fn to_pedersen_commitment(&self) -> RistrettoPoint {
+        PedersenGens::default().commit(Scalar::from(self.amount), self.blinding_factor)
+    }
+
+    pub fn to_pedersen_commitment_compressed(&self) -> CompressedRistretto {
+        self.to_pedersen_commitment().compress()
+    }
```
davidrusu, line 77: Same here, we can keep the API small and let users compress if they need to

davidrusu (dismissed):

src/dbc_content.rs
dan-da: done.

dan-da:
@davidrusu This PR was previously introducing a pattern of ::to_bytes() -> Vec<u8> and ::hash() -> Hash for many structs.   sn_payment is also doing a similar thing.    The hash function was always exactly the same, it just calls to_bytes() and hashes it.  This repeating boilerplate code seemed icky, so instead I made a trait `ByteHash` which impls a default hash() fn and requires to_bytes().   I think it is nicer, but let me know what you think.

One thing I don't have a great solution for is from_bytes(), ie for symmetry.   That would require more rigor to account for the exact sizes of every field.  And we don't need from_bytes() for generating a hash or anything else at present.   So I did not make it part of the trait.

For the most part, behavior remains the same.  One thing that changes is that dbc::spend_key_index() was using the hash of DbcContent and DbcTransaction as byte input to the hash.    Now it uses their to_bytes() instead.  I think that should be fine/equivalent, but figured it is worth noting.

dan-da:
obsoleted by ringct.   closing.

PR #121, opened 2022-02-09
integrating RingCT into sn_dbc.

dan-da:
The primary purpose and functionality of this PR is RingCt.  However, there are many other (mostly) related changes and improvements.

Highlights:

* multisig no longer supported because blst_ringct does not (yet?) support it.  Dbcs are owned by a single SecretKey
* Rng is passed to all random-using fns.  So code can be 100% deterministic.
* issuing genesis Dbc requires that mintnodes use the same deterministic seed.
* AmountSecrets is now a wrapper for blst_ringct::RevealedCommitment
* A mock spentbook is included which generates SpentProofShares and also can generate a random list of DecoyInput
* The MintNodes now must trust pubkey of all SpentBook (sections) in order to validate tx SpentProofs.
* Dbc is now a clent-side only data structure.
* Dbc can optionally be bearer and include the SecretKey inside them.
* Dbc have a base Owner and a derived one-time-use Owner.  the base owner is for interacting with third parties.  The one-time owner is for interacting with mint and spentbook.  The spentbook enforces that a one-time-owner pubkey may only be spent once.
* The spentbook validates Tx of incoming spends.  This serves to keep entries in the SpentBook valid/clean, and also it means that the SpentBook section(s) and Mint reissuing section are double-checking eachother.
* serde support is now behind feature flag "serdes"
* mint-repl now has a nifty reissue_auto command that can generate lots of dbcs randomly.
* There is now a GenesisBuilderMock for easily constructing mintnodes, spentbooks, and the genesis dbc.  Very useful for tests, examples, benches.


src/dbc.rs
```diff
@@ -1,96 +1,309 @@
...
+        if !self
+            .transaction
+            .outputs
+            .iter()
+            .any(|o| *o.public_key() == owner)
+        {
             Err(Error::DbcContentNotPresentInTransactionOutput)
         } else {
             Ok(())
         }
```
davidrusu, line 218: Should we also call `self.confirm_provided_amount_matches_commitment(..)`?

src/dbc.rs
```diff
@@ -186,142 +403,120 @@ mod tests {
...
 
-        let genesis_dbc = Dbc {
-            content: genesis.dbc_content,
-            transaction: genesis.transaction,
-            transaction_sigs: BTreeMap::from_iter([(
-                crate::genesis_dbc_input(),
-                (genesis_key, genesis_sig),
-            )]),
-        };
+        let input_owners: Vec<OwnerOnce> = (0..=n_inputs.coerce())
```
davidrusu, line 421: Did you mean n_inputs + 1 owners?
```suggestion
        let input_owners: Vec<OwnerOnce> = (0..n_inputs.coerce())
```

src/dbc.rs
```diff
@@ -186,142 +403,120 @@ mod tests {
...
+            .map(|_| OwnerOnce::from_owner_base(Owner::from_random_secret_key(&mut rng), &mut rng8))
```
davidrusu, line 422: Also, I've been liking `std::iter::repeat_with(|| ...).take(n)` as a replacement for `(0..n).into_iter().map(|_| ...)`.

src/dbc.rs
```diff
@@ -186,142 +403,120 @@ mod tests {
...
-                SpentProofShare {
-                    spent_sig,
-                    spentbook_pks,
-                    spentbook_sig_share,
-                },
-            );
+        for (i, (mlsag, _mlsag_material)) in reissue_tx
+            .mlsags
+            .iter()
+            .zip(material.inputs.iter())
```
davidrusu, line 477: Why zip if we are not using the `mlsag_material`?

src/dbc.rs
```diff
@@ -186,142 +403,120 @@ mod tests {
...
+        let mut fuzzed_transaction_sigs: BTreeMap<KeyImage, (PublicKey, Signature)> =
             BTreeMap::new();
 
-        // Add valid sigs
+        let mint_pk = mint_node.key_manager().public_key_set()?.public_key();
         fuzzed_transaction_sigs.extend(
             reissue_share
                 .mint_node_signatures
-                .keys()
+                .into_iter()
```
davidrusu, line 513: Seems like we don't need the values, can we keep `.keys()`?

src/dbc.rs
```diff
@@ -330,89 +525,117 @@ mod tests {
...
+                );
             }
         }
 
         // Valid mint signatures for inputs not present in the transaction
         for _ in 0..n_extra_input_sigs.coerce() {
-            fuzzed_transaction_sigs
-                .insert(rand::random::<SpendKey>(), (genesis_key, mint_sig.clone()));
+            fuzzed_transaction_sigs.insert(
+                Default::default(),
```
davidrusu, line 567: Feels a bit loose to have a `Default` impl on `KeyImage`, How about we just be explicit with a `KeyImage::random(&mut rng)` and drop the Default?

src/dbc.rs
```diff
@@ -330,89 +525,117 @@ mod tests {
...
+            .owner_once(&owner_once.owner_base().secret_key()?)?
+            .public_key_blst();
+
+        // Check if commitment in AmountSecrets matches the commitment in tx OutputProof
+        let commitments_match = dbc
+            .confirm_provided_amount_matches_commitment(
+                &owner_once.owner_base().secret_key()?,
+                &fuzzed_amt_secrets,
+            )
+            .is_ok();
```
davidrusu, line 595: If we move this confirmation into `confirm_valid()`, we should remember to remove this

src/spent_proof.rs
```diff
@@ -1,56 +1,57 @@
...
-    /// Signature from dbc.spend_key() over the transaction
-    pub spent_sig: Signature,
+    pub key_image: KeyImage,
 
     /// The Spentbook who notarized that this DBC was spent.
     pub spentbook_pks: PublicKeySet,
 
     pub spentbook_sig_share: NodeSignature,
+
+    pub public_commitments: Vec<Commitment>,
```
davidrusu, line 32: We need to be including the public_commitments in the spent proof signature. Otherwise mint's can't trust that the public commitments here were checked for validity by the spent book nodes

src/validation.rs
```diff
@@ -0,0 +1,147 @@
...
+    }
+
+    fn validate_without_sigs_internal<K: KeyManager>(
+        mint_verifier: &K,
+        transaction: &RingCtTransaction,
+        transaction_hash: Hash,
+        spent_proofs: &BTreeSet<SpentProof>,
+    ) -> Result<(), Error> {
+        if transaction.mlsags.is_empty() {
+            return Err(Error::TransactionMustHaveAnInput);
```
davidrusu, line 79: This check should be moved to `RingCtTransaction::verify()` if it's not already there.

src/validation.rs
```diff
@@ -0,0 +1,147 @@
...
+
+        // Verify that each KeyImage is unique in this tx.
+        let keyimage_unique: BTreeSet<KeyImage> = transaction
+            .mlsags
+            .iter()
+            .map(|m| m.key_image.into())
+            .collect();
+        if keyimage_unique.len() != transaction.mlsags.len() {
+            return Err(Error::KeyImageNotUniqueAcrossInputs);
+        }
```
davidrusu, line 92: This check should be moved to `RingCtTransaction::verify()` if it's not already there.

src/validation.rs
```diff
@@ -0,0 +1,147 @@
...
+
+        // Verify that each pubkey is unique in this transaction.
+        let pubkey_unique: BTreeSet<KeyImage> = transaction
+            .outputs
+            .iter()
+            .map(|o| (*o.public_key()).into())
+            .collect();
+        if pubkey_unique.len() != transaction.outputs.len() {
+            return Err(Error::PublicKeyNotUniqueAcrossOutputs);
+        }
```
davidrusu, line 102: This check should be moved to `RingCtTransaction::verify()` if it's not already there.

davidrusu (commented):
Left a few comments that we need to fix, but it's looking good.

examples/mint-repl/mint-repl.rs
```diff
@@ -878,62 +773,160 @@ fn prepare_reissue_cli(mintinfo: &mut MintInfo) -> Result<()> {
...
+    let mut rng8 = rand8::thread_rng();
+
+    let num_reissues: usize =
+        readline_prompt_nl_default("\nHow many reissues to perform [10]: ", "10")?.parse()?;
+
+    for _i in 1..=num_reissues {
+        let max_inputs = vec![mintinfo.reissue_auto.unspent_dbcs.len(), 10]
+            .into_iter()
+            .min()
+            .unwrap_or(10);
```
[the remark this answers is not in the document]
dan-da: done.

examples/mint-repl/mint-repl.rs
```diff
@@ -878,62 +773,160 @@ fn prepare_reissue_cli(mintinfo: &mut MintInfo) -> Result<()> {
...
+
+        let mut tx_builder = TransactionBuilder::default();
+
+        for dbc in input_dbcs.iter() {
+            let base_sk = dbc.owner_base().secret_key()?;
+            let decoy_inputs = mintinfo.spentbook()?.random_decoys(STD_DECOYS, &mut rng8);
+
+            tx_builder = tx_builder.add_input_dbc(dbc, &base_sk, decoy_inputs, &mut rng8)?;
+
+            mintinfo.reissue_auto.unspent_dbcs.remove(&dbc.hash());
```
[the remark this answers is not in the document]
dan-da: done.

examples/mint-repl/mint-repl.rs
```diff
@@ -878,62 +773,160 @@ fn prepare_reissue_cli(mintinfo: &mut MintInfo) -> Result<()> {
...
+        }
+
+        let inputs_sum = tx_builder.inputs_amount_sum();
+
+        while tx_builder.outputs_amount_sum() < inputs_sum {
+            // randomize output amount
+            let diff = inputs_sum - tx_builder.outputs_amount_sum();
+            let amount = if diff == 1 { 1 } else { rng.gen_range(1, diff) };
```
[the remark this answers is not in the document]
dan-da: done.  though I had to modify the while loop condition like so:
```
while tx_builder.outputs_amount_sum() < inputs_sum || tx_builder.outputs().is_empty() {
```
otherwise zero valued outputs can become inputs which sum to zero, so the while body does not exec, leading to an invalid tx with zero outputs.

that's a very subtle one.

examples/mint-repl/mint-repl.rs
```diff
@@ -878,62 +773,160 @@ fn prepare_reissue_cli(mintinfo: &mut MintInfo) -> Result<()> {
...
+                Output {
+                    amount,
+                    public_key: owner_once.as_owner().public_key_blst(),
+                },
+                owner_once,
+            );
+        }
+
+        let (reissue_tx, revealed_commitments, _material, output_owners) =
+            tx_builder.build(&mut rng8)?;
```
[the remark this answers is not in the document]
dan-da: created #122 for this.

examples/mint-repl/mint-repl.rs
```diff
@@ -878,62 +773,160 @@ fn prepare_reissue_cli(mintinfo: &mut MintInfo) -> Result<()> {
...
 }
 
 /// Performs reissue
-fn reissue(
-    mintinfo: &mut MintInfo,
-    reissue_request: &ReissueRequest,
-    output_pk_pks: &HashMap<PublicKey, PublicKeySet>,
-) -> Result<()> {
-    let mut dbc_builder = DbcBuilder::new(reissue_request.transaction.clone());
+fn reissue(mintinfo: &mut MintInfo, reissue_request: ReissueRequestUnblinded) -> Result<()> {
```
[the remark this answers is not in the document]
dan-da: done.   /s/Unblinded/Revealed/g

src/amount_secrets.rs
```diff
@@ -0,0 +1,203 @@
...
+        self.0.blinding
+    }
+
+    /// Convert to bytes
+    pub fn to_bytes(&self) -> Vec<u8> {
+        self.0.to_bytes()
+    }
+
+    /// build AmountSecrets from fixed size byte array.
+    pub fn from_bytes(bytes: [u8; AMT_SIZE + BF_SIZE]) -> Self {
```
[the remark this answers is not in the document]
dan-da: created #123 for this.

src/amount_secrets.rs
```diff
@@ -0,0 +1,203 @@
...
+        .unwrap();
+
+        Self(RevealedCommitment {
+            value: amount,
+            blinding: blinding_factor,
+        })
+    }
+
+    /// build AmountSecrets from byte array reference
+    pub fn from_bytes_ref(bytes: &[u8]) -> Result<Self, Error> {
```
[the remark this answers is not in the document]
dan-da: not really.  I forget the details, but there was some ugliness that made me think its better just to keep to_bytes and from_bytes as regular impl methods.

we may want to revisit later, to make standard to/from bytes and also hash() across structs and crates.

src/blst.rs
```diff
@@ -0,0 +1,131 @@
...
+
+impl AsRef<G1Affine> for KeyImage {
+    fn as_ref(&self) -> &G1Affine {
+        &self.0
+    }
+}
+
+impl Eq for PublicKeyBlstMappable {}
+
+impl PartialOrd for KeyImage {
```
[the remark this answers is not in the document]
dan-da: gotta keep clippy happy.

src/blst.rs
```diff
@@ -0,0 +1,131 @@
...
+    }
+}
+
+impl Ord for PublicKeyBlstMappable {
+    fn cmp(&self, other: &Self) -> Ordering {
+        self.0.to_compressed().cmp(&other.0.to_compressed())
+    }
+}
+
+impl Hash for PublicKeyBlstMappable {
```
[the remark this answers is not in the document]
dan-da: yep, you got it.

src/builder.rs
```diff
@@ -1,144 +1,275 @@
...
-/// Builds a ReissueRequest from a ReissueTransaction and
+/// Builds a ReissueRequest from a RingCtTransaction and
 /// any number of (input) DBC spent proof shares.
+#[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
 #[derive(Debug)]
 pub struct ReissueRequestBuilder {
-    pub reissue_transaction: ReissueTransaction,
-    pub spent_proof_shares: BTreeMap<SpendKey, HashSet<SpentProofShare>>,
+    pub transaction: RingCtTransaction,
+    pub spent_proof_shares: BTreeMap<usize, HashSet<SpentProofShare>>,
```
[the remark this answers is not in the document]
dan-da: This is done because the commitments must be in original input order for this code in TransactionValidator to work:
```
        let public_commitments: Vec<Vec<Commitment>> = spent_proofs_sorted
            .iter()
            .map(|s| s.public_commitments.clone())
            .collect();

        transaction.verify(&public_commitments)?;
```
Looking at it again, I should be able to get them in proper order by matching with the tx.mlsags inputs, which are ordered.   anyway, it's a big enough change I will do it in a separate commit, 2nd pass.

src/builder.rs
```diff
@@ -1,144 +1,275 @@
...
                     .next()
-                    .ok_or(Error::ReissueRequestMissingSpentProofShare(*spend_key))?;
+                    .ok_or(Error::ReissueRequestMissingSpentProofShare(*input_index))?;
 
-                if shares
+                if !shares
                     .iter()
                     .map(SpentProofShare::spentbook_pks)
-                    .any(|pks| pks != any_share.spentbook_pks())
+                    .any(|pks| pks == any_share.spentbook_pks())
```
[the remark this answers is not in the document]
dan-da: done. nice catch.

src/builder.rs
```diff
@@ -1,144 +1,275 @@
...
                 }
 
-                let spent_sig = any_share.spent_sig.clone();
+                if !shares
+                    .iter()
+                    .map(|s| &s.public_commitments)
+                    .any(|pc| *pc == any_share.public_commitments)
+                {
+                    return Err(Error::ReissueRequestPublicCommitmentMismatch);
+                }
```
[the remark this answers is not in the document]
dan-da: done. nice catch.

re future work:  perhaps.  hopefully we can make it possible for client to call RingCtTransaction::verify() before sending tx to spentbook or mint.  If we can do that, maybe we don't need any other build validations...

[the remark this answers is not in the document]
dan-da: opened #124

src/builder.rs
```diff
@@ -224,19 +365,18 @@ impl DbcBuilder {
...
             }
 
             // Verify that each input has a NodeSignature
-            for input in self.reissue_transaction.inputs.iter() {
-                if rs.mint_node_signatures.get(&input.spend_key()).is_none() {
+            for mlsag in rs.transaction.mlsags.iter() {
+                if !rs
+                    .mint_node_signatures
+                    .keys()
+                    .any(|k| *k == mlsag.key_image.into())
```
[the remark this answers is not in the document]
dan-da: done.

src/builder.rs
```diff
@@ -258,37 +398,349 @@ impl DbcBuilder {
...
+            let mut genesis_set: Vec<GenesisDbcShare> = Default::default();
+            let mut mint_nodes: Vec<MintNode<SimpleKeyManager>> = Default::default();
+            let mut spentbook_nodes: Vec<SpentBookNodeMock> = Default::default();
+
+            for mint_node in self.mint_nodes.into_iter() {
+                // Each MintNode must start from same random seed when issuing Genesis.
+                let mut rng8 = rng8_initial.clone();
+
+                let (mut mint_node, genesis_dbc_share) =
+                    mint_node.issue_genesis_dbc(self.genesis_amount, &mut rng8)?;
```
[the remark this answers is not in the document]
dan-da: opened #125

src/dbc.rs
```diff
@@ -1,96 +1,309 @@
...
+/// The included Owner is called an Owner Base.  The public key can be
+/// given out to multiple parties and thus multiple Dbc can share
+/// the same Owner Base.
+///
+/// The Mint and Spentbook never see the Owner Base.  Instead, when a
+/// transaction Output is created for a given Owner Base, a random derivation
+/// index is generated and used to derive a one-time-use Owner Once.
+///
+/// The Owner Once is used for a single transaction only and must be unique
+/// within a transaction.
```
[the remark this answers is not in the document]
dan-da: changed to:
> within the transaction as well as globally for the output DBC's to be spendable.

src/dbc.rs
```diff
@@ -1,96 +1,309 @@
...
+/// from the user (owned).  Subsequent code is then the same for both
+/// types.
+#[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
+#[derive(Debug, Clone)]
 pub struct Dbc {
     pub content: DbcContent,
-    pub transaction: DbcTransaction,
-    pub transaction_sigs: BTreeMap<SpendKey, (PublicKey, Signature)>,
+    pub transaction: RingCtTransaction,
+    pub transaction_sigs: BTreeMap<KeyImage, (PublicKey, Signature)>,
```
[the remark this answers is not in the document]
dan-da: done.  more diffs, yay!

src/validation.rs
dan-da: created #126

src/dbc.rs
```diff
@@ -1,96 +1,309 @@
...
+
+    /// decypts and returns the AmountSecrets
+    /// will return an error if the SecretKey is not available.  (not bearer)
+    pub fn amount_secrets_bearer(&self) -> Result<AmountSecrets> {
+        self.amount_secrets(&self.owner_base().secret_key()?)
+    }
+
+    /// returns KeyImage for the owner's derived public key
+    /// This is useful for checking if a Dbc has been spent.
+    pub fn key_image(&self, base_sk: &SecretKey) -> Result<KeyImage> {
```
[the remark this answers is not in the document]
dan-da: created #126

src/dbc.rs
dan-da: maybe, but that would be "invalid" in a different sense than other confirm_valid() errors.

Presently we can distinguish between:
1. Dbc is valid and amount secrets are valid
2. Dbc is valid and amount secrets are invalid
3. Dbc is invalid and amount secrete are invalid
4. Dbc is invalid and amount secrets are valid

Of these, I think 1 and 2 are most important.  Not much need to distinguish between 3 and 4 usually I guess.

One way to retain same granularity would be to return an Enum from confirm_valid() instead of just ().

Or maybe we just keep the present API.

Anyway, I will open an issue for this.

dan-da: created #127

src/dbc.rs
dan-da: yeah, that works.

src/dbc.rs
dan-da: interesting.  I think I will leave it as-is for now though.   it's actually 1 extra line to do it the repeat_with() way, after cargo fmt.

```
        let input_owners: Vec<OwnerOnce> = std::iter::repeat_with(|| {
            OwnerOnce::from_owner_base(Owner::from_random_secret_key(&mut rng), &mut rng8)
        })
        .take(n_inputs.coerce())
        .collect();
```
vs.
```
        let input_owners: Vec<OwnerOnce> = (0..n_inputs.coerce())
            .map(|_| OwnerOnce::from_owner_base(Owner::from_random_secret_key(&mut rng), &mut rng8))
            .collect();
```

src/dbc.rs
dan-da: fixed.

src/dbc.rs
dan-da: the into_iter() avoided the need to clone() each in_owner inside the map().

I've changed it to use into_keys() instead.

src/dbc.rs
dan-da: done.   added ::random(), and removed derive Default.

```
    pub fn random(rng: &mut impl rand8::RngCore) -> Self {
        Self(G1Projective::random(rng).to_affine())
    }
```

src/dbc.rs
dan-da: commented in #127

src/spent_proof.rs
dan-da: ok, this is a bit more involved and I think requires mods in spentbook.rs and also validation.rs.    leaving for tmw.

src/validation.rs
dan-da: created #128

src/validation.rs
dan-da: created #130

src/spent_proof.rs
dan-da: ok, I made a SpentProofContent to hold all the data that SpentBook needs to sign, and it now holds the transaction hash.
```
pub struct SpentProofContent {
    pub key_image: KeyImage,
    pub transaction_hash: Hash,
    pub public_commitments: Vec<Commitment>,
}
```
So `SpentProofShare` and `SpentProof` both hold a `SpentProofContent` plus signing info.

confirm_valid() also verifies that the incoming tx hash matches this one.  And did some cleanup/refactoring in spent_proof.rs.  So I think it's good now.

commit 59d7f137bb5275b50dd7c6c4aedd06ee5cc85702

src/builder.rs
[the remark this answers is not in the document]
dan-da: done.  in 08e5883755c5707d4be626f494bcd60d9e89abfa

src/validation.rs
dan-da: added comment in #128

davidrusu (approved):
Looks good!

PR #133, opened 2022-02-17
remove Key image maths form sn_dbc #126

### LINKED ISSUE #126: move Dbc::key_image() into ringct crate.
> This function should really be defined in blst-ringct. We can keep the `key_image` method here, but it should delegate to a `blst_ringct::key_image(..)` for the key-image math

_Originally posted by @davidrusu in https://github.com/maidsafe/sn_dbc/pull/121#discussion_r805183106_

grumbach:
- fixes #126


davidrusu (approved):
:ship:

Issue #125, opened 2022-02-14
hard-code genesis dbc RingCtMaterial


dan-da:
Given that we want all nodes to come up with the same genesis dbc, (that's why we're cloning the rng above), how about we pass in a pre-agreed upon RingCTMaterial with all random values already set. This could be generated ahead of time and hard coded into our dbc system.

If you agree, lets do this as a seperate PR as well. I think there will be some interesting things to discuss on how to properly seed the genesis DBC

_Originally posted by @davidrusu in https://github.com/maidsafe/sn_dbc/pull/121#discussion_r805182585_

dan-da:
addressed by #135

PR #135, opened 2022-02-18
Genesis material pr

dan-da:
In a nutshell this PR:
* makes the genesis amount fixed, so GenesisMaterial can be known/used by any node without coordination.  (mints, spentbooks, wallets, auditors, etc)
* sets the genesis amount to u64::MAX, enabling full use of u64 space for maximum divisibility
* removes GenesisDbcShare and MintNode::issue_genesis_dbc()
* sets secret key and blinding values to "fun" values.  We can consider other possibilities.
* updates dependent code as necessary.


dan-da:
addresses #125

src/builder.rs
```diff
@@ -607,123 +597,85 @@ pub mod mock {
...
         )> {
-            let mut spent_proof_shares: Vec<SpentProofShare> = Default::default();
-            let mut genesis_set: Vec<GenesisDbcShare> = Default::default();
             let mut mint_nodes: Vec<MintNode<SimpleKeyManager>> = Default::default();
-            let mut spentbook_nodes: Vec<SpentBookNodeMock> = Default::default();
 
-            for mint_node in self.mint_nodes.into_iter() {
-                // Each MintNode must start from same random seed when issuing Genesis.
-                let mut rng8 = rng8_initial.clone();
+            // note: rng is necessary for RingCtMaterial::sign().
```
davidrusu, line 611: I skimmed the blst-ringct / blst-bulletproofs crates to see how they use `rng` in signing, I think it's safe to say that we don't really need deterministic values here, the rng is used in the rangeproofs and in the pseudo-commitements.

So, even though it would be nice to have a deterministic `sign()` method, it shouldn't matter which RNG is used in practice to build the genesis as everyone can unblind the values and see that the genesis was build using the correct amounts.

src/mint.rs
```diff
@@ -86,80 +67,6 @@ impl<K: KeyManager> MintNode<K> {
...
-                dbc_content,
-                amount_secrets,
-                owner_once,
-                transaction,
-                transaction_sig,
-                input_key_image,
-            },
-        ))
-    }
-
```
davidrusu, line 162: Nice cleanup :)

davidrusu (dismissed):
:rocket:

benches/reissue.rs
```diff
@@ -170,6 +170,53 @@ fn bench_reissue_100_to_1(c: &mut Criterion) {
     });
 }
 
+fn generate_dbc_of_value(
+    amount: Amount,
+    rng: &mut impl rand::RngCore,
+    rng8: &mut (impl rand8::RngCore + rand_core::CryptoRng),
+) -> Result<(MintNode<SimpleKeyManager>, SpentBookNodeMock, Dbc)> {
+    let (mint_node, mut spentbook_node, genesis_dbc, _genesis_material, _amount_secrets) =
+        GenesisBuilderMock::init_genesis_single(rng, rng8)?;
```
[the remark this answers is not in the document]
dan-da: super annoying.

PR #136, opened 2022-02-18
fix: change BTreeSet<SpentProofShare> to HashSet

dan-da:
bug: The ReissueRequestBuilder was only storing 1 SpentProofShare per
KeyImage because the SpentProofShare were being ordered by KeyImage
which is the same for all items in the set for a given input.

The fix is to use a HashSet instead.

changes:

* use HashSet<SpentProofShare> instead of BTreeSet in ReissueRequestBuilder
* add NodeSignature::to_bytes()
* impl (PartialEq, Eq) manually on SpentProofShare
* remove Ord, PartialOrd impls on SpentProofShare
* impl Hash on SpentProofShare
* add SpentProofShare::to_bytes()

### DIFF
--- a/src/spent_proof.rs
+++ b/src/spent_proof.rs
@@ -64,7 +64,7 @@ impl Ord for SpentProofContent {
 /// A share of a SpentProof, combine enough of these to form a
 /// SpentProof.
 #[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
-#[derive(Debug, Clone, PartialEq, Eq)]
+#[derive(Debug, Clone)]
 pub struct SpentProofShare {
     /// data to be signed
     pub content: SpentProofContent,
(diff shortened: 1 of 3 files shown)


src/spent_proof.rs
```diff
@@ -74,15 +74,21 @@ pub struct SpentProofShare {
     pub spentbook_sig_share: NodeSignature,
 }
 
-impl PartialOrd for SpentProofShare {
-    fn partial_cmp(&self, other: &Self) -> Option<Ordering> {
-        Some(self.cmp(other))
+// impl manually to avoid clippy complaint about Hash conflict.
```
davidrusu, line 77: It'd be better if we can derive both PartialEq, Eq, and Hash. Was one of the fields not hasheable?

dan-da: yes, Hash cannot be derived for SpentProofContent because it includes a Vec<Commitment>, where Commitment is typedef for blstrs::G1Affine, which does not impl Hash.   and of course blstrs is outside our control.

We do have PublicKeyBlstMappable in blst.rs which is a newtype wrapper for G1Affine and impls Hash.  So if we typedef'd Commitment to PublicKeyBlstMappable, then it could be made to work, but is a larger change because any code that uses the Vec<Commitment> has to be modified (builder, spentbook, validation)... I just tried it and its a bit involved to get those building again.

So I think for now let's just go with this.  Hopefully can find some time to clean it up soonish.

davidrusu (approved):

Issue #128, opened 2022-02-15
This check should be moved to `RingCtTransaction::verify()` if it's not already there.


dan-da:
> This check should be moved to `RingCtTransaction::verify()` if it's not already there.

_Originally posted by @davidrusu in https://github.com/maidsafe/sn_dbc/pull/121#discussion_r805207914_

dan-da:
and this one.  https://github.com/maidsafe/sn_dbc/pull/121#discussion_r805207975

dan-da:
fixed in #137

PR #137, opened 2022-02-21
feat: moved input and key image uniqueness checks to blst-ringct

### LINKED ISSUE #129: This check should be moved to `RingCtTransaction::verify()` if it's not already there.
> This check should be moved to `RingCtTransaction::verify()` if it's not already there.

_Originally posted by @davidrusu in https://github.com/maidsafe/sn_dbc/pull/121#discussion_r805207975_

grumbach:
- Fixes #129
- and #128

### DIFF
--- a/src/error.rs
+++ b/src/error.rs
@@ -51,9 +51,6 @@ pub enum Error {
     #[error("OutputProof not found in transaction outputs")]
     OutputProofNotFound,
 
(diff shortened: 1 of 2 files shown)


davidrusu (approved):
Make sure to bump our blst-ringct dependency to pull in your changes there

PR #138, opened 2022-02-22
fix: return Spentbook errors. don't panic

dan-da:
Adds Error variants for a couple of spentbook errors.

It's a bit gross to put them sn_dbc::Error, however I'm thinking it will be split up soon when we move to a modules approach for MintNode, Wallet, Spentbook.  So this is temporary.

* fix: return Error::SpentbookKeyImageAlreadySpent instead of panic
* fix: return Error::SpentbookRingSizeMismatch instead of assert failure
* add Error::SpentbookKeyImageAlreadySpent
* add Error::SpentbookRingSizeMismatch

### DIFF
--- a/src/spentbook.rs
+++ b/src/spentbook.rs
@@ -149,17 +149,24 @@ impl SpentBookNodeMock {
                             })
                             .collect();
 
+                        if output_proofs.len() != mlsag.public_keys().len() {
+                            return Err(crate::Error::SpentbookRingSizeMismatch(
+                                mlsag.public_keys().len(),
+                                output_proofs.len(),
+                            ));
+                        }
+
                         // collect commitments from OutputProofs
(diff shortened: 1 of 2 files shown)


src/spentbook.rs
```diff
@@ -214,7 +221,7 @@ impl SpentBookNodeMock {
             })
         } else {
             // fixme: return an error.  can wait until we refactor into a Mock feature flag.
```
davidrusu, line 223: rm this fixme?

davidrusu (approved):
Looks good

src/spentbook.rs
dan-da: will remove it in coming days while doing pass of fixme/todo/unwraps.  polishing, etc.

PR #150, opened 2022-03-02
feat: verify (always) Amount matches Commitment

dan-da:
High level changes:

1. The check that secret amount matches commitment is now moved inside
Dbc::verify() so there is no chance of a wallet forgetting to do it.

2. We standardize on verify/verification instead of validate/validation.
This matches usage in ringct and blsttc, so it makes the api more
regular.

Details:

* change validate command to verify in mint-repl
* impl From<AmountSecrets> instead of Into<RevealedCommitment>
* impl From<AmountSecrets> instead of Into<Amount>
* replace validate/validation with verify/verification in code/comments
* rename validation.rs to verification.rs
* rename TransactionValidator to TransactionVerifier
* rename Dbc::confirm_valid() to verify()
* verify that secret Amount matches public Commitment inside Dbc::verify()
* update tests appropriately

### DIFF
--- a/src/dbc.rs
+++ b/src/dbc.rs
@@ -6,7 +6,7 @@
 // KIND, either express or implied. Please review the Licences for the specific language governing
 // permissions and limitations relating to use of the SAFE Network Software.
 
-use crate::{AmountSecrets, KeyImage, Owner, SpentProof, TransactionValidator};
+use crate::{AmountSecrets, KeyImage, Owner, SpentProof, TransactionVerifier};
 use crate::{DbcContent, DerivationIndex, Error, KeyManager, Result};
 use blst_ringct::ringct::{OutputProof, RingCtTransaction};
 use blst_ringct::{RevealedCommitment, TrueInput};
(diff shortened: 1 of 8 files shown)


src/dbc.rs
```diff
@@ -241,34 +241,17 @@ impl Dbc {
     /// If the merchant were to send the goods without first performing
     /// this check, then they could be stuck with an unspendable Dbc
     /// and no recourse.
-    pub fn confirm_provided_amount_matches_commitment(
-        &self,
-        base_sk: &SecretKey,
-        amount: &AmountSecrets,
-    ) -> Result<()> {
+    pub(crate) fn confirm_amount_matches_commitment(&self, base_sk: &SecretKey) -> Result<()> {
```
davidrusu, line 244:
```suggestion
    pub(crate) fn verify_amount_matches_commitment(&self, base_sk: &SecretKey) -> Result<()> {
```

davidrusu (dismissed):
LGTM

PR #152, opened 2022-03-04
Builder verification pr

dan-da:
fix: verify Tx in DbcBuilder::build()

Previously DbcBuilder::build() was only doing some incomplete
validations.  This commit verifies the Tx which also requires
that DbcBuilder::build() accept a KeyManager param, which has
mintnode/spentbooknode trust implications/requirements that
are now documented.

* doc: add doc comment for TransactionVerifier::verify() and
        ::verify_without_sigs()
* rename mint_verifier param to verifier
* doc: add doc comment for Dbc::verify()
* optimization: iter/collect() mint sigs once per tx, not per Dbc
* add KeyManager arg to DbcBuilder::build()
* call TransactionVerifier::verify() within DbcBuilder::build()
* adapt tests, examples, benches to pass key_manager into ::build()

------
fix: avoid unnecessary sig verifications

Previously we were performing duplicate signature verifications.

This occurred whenever two or more inputs to our tx (b) were each outputs of the
same source tx (a). In this case, KeyManager::verify() was being called using
the exact same tx_hash, mint_pk, and mint_sig for each input.

The change is to:
1. separate logic that verifies a mint sig is provided for each
tx from the logic that actually verifies each sig.
2. generate a list of unique pk+sig.
3. verify tx hash against each pk+sig.

### DIFF
--- a/benches/reissue.rs
+++ b/benches/reissue.rs
@@ -108,7 +108,7 @@ fn bench_reissue_100_to_1(c: &mut Criterion) {
     let reissue_share = mintnode.reissue(rr).unwrap();
 
     dbc_builder = dbc_builder.add_reissue_share(reissue_share);
-    let dbcs = dbc_builder.build().unwrap();
+    let dbcs = dbc_builder.build(mintnode.key_manager()).unwrap();
 
     let output_owner_once =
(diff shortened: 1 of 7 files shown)


davidrusu (approved):
LGTM

PR #155, opened 2022-03-11
fix: verify that ReissueShare match

dan-da:
We modify ReissueShare to facilitate verifying that common fields are the same between all mint nodes. (All fields except sig-share must match.)

We also remove from ReissueShare an unnecessary Vec of redundant MintNodeSignature keyed by input KeyImage.  All MintNodeSignature were signing the same Tx, so only one copy is needed.

The above changes enable DbcBuilder::build() to be simplified quite a bit, and it is more correct because now it verifies that all common fields of all ReissueShare are equal.

While we are at it, we do a bit of cleanup/normalization of the API such as renaming NodeSignature to IndexedSignatureShare.

* remove: MintNodeSignature
* remove: MintNodeSignatures
* modify ReissueShare fields
* NodeSignature --> IndexedSignatureShare
* add ReissueShare::to_common_bytes()
* fix: DbcBuilder now correctly checks that ReissueShares match
* Error::ReissueSharePublicKeySetMismatch --> ReissueShareMismatch
* update tests

### DIFF
--- a/src/mint.rs
+++ b/src/mint.rs
@@ -14,17 +14,14 @@
 // Outputs <= input value
 
 use crate::{
-    Error, Hash, KeyImage, KeyManager, NodeSignature, PublicKey, PublicKeySet, Result, SpentProof,
-    SpentProofShare, TransactionVerifier,
+    Error, Hash, IndexedSignatureShare, KeyImage, KeyManager, PublicKey, PublicKeySet, Result,
+    SpentProof, SpentProofShare, TransactionVerifier,
 };
 use blst_ringct::ringct::RingCtTransaction;
-use std::collections::{BTreeMap, BTreeSet};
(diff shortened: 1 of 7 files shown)


src/mint.rs
```diff
@@ -38,7 +35,25 @@ pub struct ReissueRequest {
 pub struct ReissueShare {
     pub transaction: RingCtTransaction,
     pub spent_proofs: BTreeSet<SpentProof>,
-    pub mint_node_signatures: MintNodeSignatures,
+    pub mint_public_key_set: PublicKeySet,
+    pub mint_signature_share: IndexedSignatureShare,
```
davidrusu, line 39: much better

davidrusu (approved):
Nice cleanup

PR #157, opened 2022-03-16
integrate blsttc/blstrs and remove mint::*

dan-da:
This PR includes and obsoletes #154.  removes 827 LOC.   :+1: 

Alternatively, if #154 were merged as-is, I could rebase this branch to include only the mint removal.

-----

**integrate blsttc/blstrs**

highlights:

* we only use blsttc keys, no more Scalar/G1Affine.
* Amount is now defined in ringct and re-exported in sn_dbc.
* get rid of rand7, now rand8 everywhere
* use ringct::Output::new() and TrueInput::new()
* gets rid of BlsHelper, SecretKeyBlst, PublicKeyBlst
* TxBuilder methods to add output by Amount+OwnerOnce, simplifying tx creation.

lowlights:

* depends on lots of forks/branches in various repos and they all need to be merged/published.

------

remove mint::*

feat: remove mint nodes and related data structures

We realized that mint nodes no longer serve a useful purpose
as the spentbook nodes now both verify and sign the transaction.

By removing the mint nodes:
1. The client has less work to do.
2. Safe network elder nodes need only have a single role and do less work.
3. we remove a bunch of code from sn_dbc, making remaining code easier to learn/audit/maintain.
4. the system becomes more flexible as a spentbook can be anything that can generate a SpentProof.

* bench TransactionVerifier::verify() instead of MintNode::reissue()
* remove mint::* from mint-repl
* make TransactionBuilder::build() return only a DbcBuilder
* remove ReissueRequestBuilder
* modify DbcBuilder to build using SpentProofs
* add DbcBuilder::spent_proofs()
* remove all 'mint', 'Mint' from code/comments.
* remove mint sigs from Dbc
* fuzz spent proofs in prop_dbc_verification()
* remove obsolete/unused error variants
* remove all lib code from mint.rs.  (tests remain there for now)
* add KeyManager error as string to Error:InvalidSpentProofSignature
* remove mint sig verification from TransactionVerifier::verify()

PR #154, opened 2022-03-07
Blsttc blstrs integration pr

dan-da:
This PR utilizes blstrs impl of blsttc.

highlights:
* we only use blsttc keys, no more Scalar/G1Affine.
* Amount is now defined in ringct and re-exported in sn_dbc.
* get rid of rand7, now rand8 everywhere
* use ringct::Output::new() and TrueInput::new()
* gets rid of BlsHelper, SecretKeyBlst, PublicKeyBlst
* TxBuilder methods to add output by Amount+OwnerOnce, simplifying tx creation.

lowlights:
* depends on lots of forks/branches in various repos and they all need to be merged/published.

PR #158, opened 2022-03-20
Decouple deps

dan-da:
<strike>note: this PR includes #154 and #157 which have not yet been merged.   If/when those are merged, this can be rebased to make a smaller diff for review.</strike>  edit: merged. this is ready for review.

----------

    feat: decouple deps and remove bls_dkg
    
    -- Change 1: re-export blsttc, blst_ringct --
    
    The idea is to give API consumers the option to use the exported crates
    without needing explicit Cargo.toml entries.
    
    This in turn decouples dependencies and reduces conflicts.
    
    See rationale/discussion here:
    https://github.com/rust-lang/api-guidelines/discussions/176
    
    -- Change 2: use deps from within blsttc, blst_ringct --
    
    blsttc and blst_ringct now reexport deps used in their public APIs.
    
    So we remove those deps from our Cargo.toml and 'use' the sub-deps
    in our code.
    
    One gross thing remains which is that both crates use rand and
    they could differ.  (although at present they are the same.)
    
    This means that sn_dbc APIs that take Rng arg to be passed to
    blsttc API can potentially be different than those that take
    an Rng arg to be passed to blst_ringct API.  Although the compiler
    presently let's us pass the same rng to both.  It's a footgun!
    
    For now, I've created an 'rng' module to make it easy to pass the
    appropriate rng, and also I updated the test cases, example, bench
    to use this methodology, to demonstrate best practice.

    Perhaps/probably a better long-term approach would be to have
    blst_ringct integrate/depend/use blsttc, so that sn_dbc has only
    a single dep and is not trying to resolve the matter.
    
    -- Change 3: remove bls_dkg dependency/feature --
    
    bls_dkg was only being used for a single function used by
    test cases.  It was simple to replace this function with
    blsttc calls which simplifies things a lot.
    
    Cargo changes:
    
    * use publshed quickcheck 1.0.3
    * use dan-da/blsttc/reexport_pr
    * remove rand, rand_core deps
    * remove xor_name dep
    * remove bls_dkg dep
    * remove dkg feature
    * remove blstrs dep
    * use single-line format for all deps. (more concise)
    
    lib.rs changes:
    * re-export blst_ringct and blsttc
    * export additional types used by builder module public API.
    * add rng module, to simplfy rng usage for callers.
    * modify bls_dkg_id() to use blsttc directly, without bls_dkg

    Code changes:
    
    * update mint-repl and reissue bench to use updated sn_dbc API.
    * update tests to separate ringct rng from blsttc rng.
    * update SimpleSigner From impl now that bls_dks is removed.

-------

edit:  Change 4: export and use a single `rand`

Our public API uses/requires rand quite prominently, with several methods requiring caller to pass in an Rng.  (The same dep issues occur for blstrs as well, but it is only indirectly exposed in sn_dbc public API).

Both blst_ringct and blsttc export rand, but they do not depend on eachother, so it is possible that they could be (in the future perhaps) using incompatible versions.  So at first glance, the cleanest seeming thing is for fn that utilize blsttc::rand to accept param of that type and fn that accept blst_ringct::rand to accept param of that type.

That is what I first impl'd but it makes using our API ugly, as sometimes callers must create both rng types in the same fn.  Worse, the compiler does not even enforce the types or issue any warning because it sees that they are *presently* using the exact same rand crate.  So it is a bit of a footgun waiting to happen for API users. I modified all the tests, examples, bench to differentiate between the two as "best practice" and found the result dis-pleasing.  We had simply gone from instantiating rng7+rng8 to rng_ringct+rng_ttc.

I made an (aborted) attempt to unify the two rand's (and blstrs) by adding a blsttc dep inside blst-bulletproofs.  So then bulletproofs uses the rand, blstrs (and I think group) from blsttc, even though it doesn't use blsttc itself.  and blst_ringct uses them from bulletproofs.  However (a) it was gross to depend on blsttc without actually using its API, and (b) it didn't truly solve the problem because from sn_dbc's perspective, there is still blst_ringct::rand and blst_ringct::bulletproofs::blsttc::rand, which could theoretically differ in the future.  So I reverted that bulletproofs commit.  (but, it remains an option.)

So finally I settled on the "solution" in https://github.com/maidsafe/sn_dbc/commit/1986cd3ba127d713d42164cef083b1423ca5efac that sn_dbc arbitrarily picks blst_ringct::rand as **the** sn_dbc rand and re-exports it.  All library code references crate::rand, including methods that pass Rng to blsttc APIs.  In doing this, sn_dbc is implicitly guaranteeing to our API users that rand (and blstrs) will remain compatible between blstrs and blsttc.  I think we can make that promise because maidsafe is planning to maintain both.  And foremost, callers were pretty much guaranteed to use it this way anyway, because the compiler allows it and it is non-obvious to use the stricter-but-correct approach.

If there is some way to make the compiler be more strict, then perhaps its better to revert https://github.com/maidsafe/sn_dbc/commit/1986cd3ba127d713d42164cef083b1423ca5efac and go with the strict approach.  Otherwise, I think this is best....

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -15,15 +15,15 @@ ringct-serde = [ "blst_ringct/serde" ]
 
 [dependencies]
 thiserror = "1.0.24"
-quickcheck_macros = "1"
-quickcheck = "1.0.3"
-blsttc = {git = "https://github.com/dan-da/blsttc", branch = "reexport_pr"}
-blst_ringct = {git = "https://github.com/dan-da/blst-ringct", branch = "fix_deps_pr"}
+blsttc = "5.1.2"
+blst_ringct = {git = "https://github.com/maidsafe/blst-ringct"}
 hex = "0.4.3"
 tiny-keccak = {version = "2.0.0", features = [ "sha3" ]}
 serde = {version = "1.0.133", features = [ "derive", "rc" ], optional = true}
 
 [dev-dependencies]
+quickcheck_macros = "1"
+quickcheck = "1.0.3"
 anyhow = "1.0.40"
 rustyline = "8.0.0"
 bincode = "1.3.3"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -10,58 +10,21 @@ authors = [ "MaidSafe Developers <dev@maidsafe.net>" ]
 edition = "2018"
 
 [features]
-dkg = [ "bls_dkg" ]
 serdes = [ "serde", "ringct-serde" ]
 ringct-serde = [ "blst_ringct/serde" ]
 
 [dependencies]
 thiserror = "1.0.24"
 quickcheck_macros = "1"
-rand = "0.8.0"
+quickcheck = "1.0.3"
+blsttc = {git = "https://github.com/dan-da/blsttc", branch = "reexport_pr"}
+blst_ringct = {git = "https://github.com/dan-da/blst-ringct", branch = "fix_deps_pr"}
 hex = "0.4.3"
-rand_core = "0.6.3"
-
-  [dependencies.blsttc]
-  git = "https://github.com/dan-da/blsttc"
-  branch = "sn_dbc_integration"
-
-  [dependencies.xor_name]
-  git = "https://github.com/iancoleman/xor_name"
-  branch = "remove_osrng"
-
-  [dependencies.quickcheck]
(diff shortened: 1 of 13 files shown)


davidrusu:
Will review once #154 goes in

dan-da:
be sure to read 'Change 4' comment I just added above.

davidrusu (approved):
🚢

PR #165, opened 2022-04-01
refactor: move mock code into new mock mod

dan-da:
This moves all "extra" functionality into a new
module "mock" with a corresponding feature flag
and its own Error enum.

The moved items are:
```
 GenesisBuilderMock --> mock::GenesisBuilder
 SpentbookNodeMock  --> mock::SpentbookNode
 SimpleKeyManager   --> mock::KeyManager
 SimpleSigner       --> mock::Signer
```
This makes the core library leaner, meaner and easier to audit/review/understand.

It also makes it extra clear what is mock functionality when using the lib.  eg, in test cases, mint-repl or sn_dbc_examples.

Cargo changes:
* add 'mock' feature flag
* set mock feature flag when building tests

Code changes:
* use mock::* in tests, reissue, mint-repl
* move GenesisBuilderMock into mock::GenesisBuilder
* remove crate::Error::Spentbook*
* add crate::Error::Mock, only when mock feature enabled.
* move SimpleKeyManager, SimpleSigner into mock::{KeyManager, Signer}
* expose pub crate::mock in lib.rs when feature enabled
* move SpentbookNodeMock into mock::SpentbookNode
* add mock::Error


benches/reissue.rs
```diff
@@ -146,9 +146,9 @@ fn bench_reissue_100_to_1(c: &mut Criterion) {
 fn generate_dbc_of_value(
     amount: Amount,
     rng: &mut (impl RngCore + CryptoRng),
-) -> Result<(SpentBookNodeMock, Dbc)> {
+) -> Result<(mock::SpentBookNode, Dbc)> {
```
maqi, line 149: not sure about the general plan of this `mock` and the real production code.
However, normally what we do is here we use a `normal` struct, say `SpentBookNode`
Then within the implementation of `SpentBookNode`, we use the `mock` feature flag to switch beteen the real production code and mock code.
Though currently we only have mock code ?

same comment to other similar usage.

dan-da: Here the meaning of "mock" is a little bit different.

Mock objects represent system components that are intended to be implemented outside of sn_dbc.

So for example, with SpentBook, there is a logical concept of a SpentBook Node, but sn_dbc only implements SpentProof in its public API.  So basically sn_dbc doesn't care what a SpentBook Node looks like, what it's API is, or anything.  It only cares that somehow there exists something which can produce a SpentProof.  So this is a very loose coupling... there is not even a Trait for a Spentbook.

Still, there is a need to write test cases that simulate a fully functional system.  So the mock SpentBookNode (and other mock objects) exist for this purpose, and for prototyping basic logic.   But they would never be used in production code.

dan-da:
note: cargo tests are failing.  Once #168 is merged, I will rebase this on top.

benches/reissue.rs
maqi: but there is still `mock` stuff appears in the production code? 
like in the `src/dbc.rs` file ?

If it does like what you say only for the testing, then better put it as a `test_utility` ?
like what we have for `safe_network`, where there is `utility` under the `tests` folder

davidrusu: @maqi Looks like all the `mock` usages in `src/dbc.rs` etc. are just in test cases. (we have unit tests inline in the source files).

The mock's are also used in the examples and benches as well, e.g. the mint-repl and reissue benchmarks makes heavy use of the mocks

davidrusu (dismissed):
LGTM

dan-da:
rebased on latest master.   I will merge it once CI tests pass.

PR #167, opened 2022-04-04
build: disable bench on windows

dan-da:
This temporarily disables building pprof dep and benches on windows
because pprof crate doesn't build on windows.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -38,7 +38,8 @@ rustyline = "8.0.0"
 bincode = "1.3.3"
 criterion = "0.3.5"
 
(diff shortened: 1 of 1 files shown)


davidrusu:
@dan-da looks like CI tests are failing on ubuntu

dan-da:
> @dan-da looks like CI tests are failing on ubuntu

yes, see #168.  It fixes an unrelated test error, and also include this patch.  Tests are passing on all platforms.

dan-da:
This was addressed in #168 which is now merged.   closing.

PR #168, opened 2022-04-05
fix: avoid true input inclusion in decoys

dan-da:
We've had some tests cases failing because blst-ringct now verifies
that every public_key referenced by a Tx input is unique.  But
the test cases were obtaining a set of (possibly duplicate) random
decoys for each input. So these decoys could duplicate eachother
and also any of them could also be a true input.

The fix is to refactor TransactionBuilder such that:
1. DecoyInputs are added to a decoy pool within the builder,
and the builder removes dups from the pool.
2. ::build() removes any true inputs from the decoys pool
and then distributes decoys from the pool amongst the inputs.
3. If there are not enough decoys in the pool then an error
is returned -- unless the 'require_all_decoys' setting is false,
in which case any remaining inputs receive no decoys.

Cargo changes:
* only build pprof and benches on unix/linux

Code changes:
* update tests, bench, mint-repl to use modified builder API
* add fields to TransactionBuilder
* add decoys_per_input field and set default to 10.
* rename TransactionBuilder::add_input_by_true_input to add_true_input()
* add API TransactionBuilder::set_decoys_per_input()
* add API TransactionBuilder::set_require_all_inputs()
* add API TransactionBuilder::add_decoy_inputs()
* modify TransactionBuilder::build() to filter true inputs from decoys
  and distribute decoys amongst inputs
* add Error::InsufficientDecoys

### DIFF
--- a/src/builder.rs
+++ b/src/builder.rs
@@ -29,15 +29,83 @@ pub type OutputOwnerMap = BTreeMap<PublicKey, OwnerOnce>;
 /// A builder to create a RingCt transaction from
 /// inputs and outputs.
 #[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
-#[derive(Debug, Default)]
+#[derive(Debug)]
 pub struct TransactionBuilder {
+    true_inputs: Vec<TrueInput>,
     ringct_material: RingCtMaterial,
     output_owner_map: OutputOwnerMap,
+    available_decoys: Vec<DecoyInput>,
+    decoys_per_input: usize,
+    require_all_decoys: bool,
+}
+
+impl Default for TransactionBuilder {
+    fn default() -> Self {
+        Self {
+            true_inputs: Default::default(),
+            ringct_material: Default::default(),
+            output_owner_map: Default::default(),
+            available_decoys: Default::default(),
+            decoys_per_input: 10, // default to 10 decoys per input.
+            require_all_decoys: true,
+        }
+    }
 }
 
 impl TransactionBuilder {
+    /// set decoys_per_input option.
+    /// allocate this many decoys to each input (from available_decoys).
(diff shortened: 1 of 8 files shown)


src/builder.rs
```diff
@@ -29,94 +29,134 @@ pub type OutputOwnerMap = BTreeMap<PublicKey, OwnerOnce>;
...
+        //   2. it requires knowing the DecoyInputs in advance, which we can't
+        //      really know until we know all the true inputs because we have
+        //      to ensure that no true input is used as a decoy input.
+        self = self.add_true_input(mlsag.true_input.clone());
         self.ringct_material.inputs.push(mlsag);
         self
     }
 
-    /// add an input given an iterator over MlsagMaterial
+    // /// add an input given an iterator over MlsagMaterial
```
davidrusu, line 113: Did you mean to add the comment here?

src/builder.rs
```diff
@@ -216,26 +248,77 @@ impl TransactionBuilder {
     }
 
     /// build a RingCtTransaction and associated secrets
-    pub fn build(self, rng: impl RngCore + CryptoRng) -> Result<DbcBuilder> {
+    pub fn build(self, rng: &mut (impl RngCore + CryptoRng)) -> Result<DbcBuilder> {
```
davidrusu, line 251: You should be able to simply make `rng` mut.
```suggestion
    pub fn build(self, mut rng: impl RngCore + CryptoRng) -> Result<DbcBuilder> {
```

Then you can replace all uses of `&mut *rng` with `&mut rng`

davidrusu (dismissed):
nice, I find the builder API gets simpler with this change too.

src/builder.rs
dan-da: fixed.

PR #177, opened 2022-04-15
chore(api): minor improvement to TransactionBuilder API

bochaco:


src/builder.rs
```diff
@@ -154,9 +154,12 @@ impl TransactionBuilder {
     }
 
     /// add an input given a list of bearer Dbcs and associated SecretKey and decoys
-    pub fn add_inputs_dbc_bearer(mut self, dbcs: impl IntoIterator<Item = Dbc>) -> Result<Self> {
-        for dbc in dbcs.into_iter() {
-            self = self.add_input_dbc_bearer(&dbc)?;
+    pub fn add_inputs_dbc_bearer<'a>(
+        mut self,
+        dbcs: impl Iterator<Item = &'a Dbc>,
```
davidrusu, line 159: Try using the `Borrow` trait so that this is usable by both borrowed and owned DBC's
```suggestion
        dbcs: impl Iterator<Item = Borrow<Dbc>>,
```

bochaco: i'll play with it, i just gave it a quick try and there are some `Sized` trait related errors....

davidrusu: I think you may have to factor out a type variable:
i.e.
```
dbcs: impl Iterator<Item = D> 
...
where D: Borrow<Dbc>
```

bochaco: yes, that seems to do it, i'll send a commit with that change there as well as in another place i found we also require moving args.

davidrusu (approved):

PR #196, opened 2022-07-07
feat!: simplifying spent proofs and TX verification public APIs

bochaco:
This is to simplify the public API exposed by sn-dbc for Tx and `SpentProofs` verification, reducing the public `Trait` to a single function, while moving the rest (`KeyManager`) to the mock as it's not required on the public API but for the example code/app.


examples/mint-repl/mint-repl.rs
```diff
@@ -85,7 +85,7 @@ fn main() -> Result<()> {
     loop {
         match rl.readline(">> ") {
             Ok(line) => {
-                let mut args = line.trim().split_whitespace();
+                let mut args = line.split_whitespace();
```
davidrusu, line 88: Does this still work when we have leading whitespace in the line input? ie. `<space>reissue`

davidrusu (approved):
q on the mint-repl change, but looks good otherwise

examples/mint-repl/mint-repl.rs
bochaco: It's supposed to, it's the linter who suggested it: https://rust-lang.github.io/rust-clippy/master/index.html#trim_split_whitespace

davidrusu: :+1: nice

PR #204, opened 2022-08-16
refactor!: reducing spent proof verification public API scope to only check it is known key

bochaco:
- Small refactor to `SpentProof` verifier public `Trait` API, reducing its scope to be a verifier of just known pub keys performing signature verification in the `SpentProof` itself before calling the `Trait`'s function.
- Minor simplification to mock key-manager and removing some unwraps from tests.

### DIFF
--- a/src/mock/genesis_builder.rs
+++ b/src/mock/genesis_builder.rs
@@ -56,15 +56,10 @@ impl GenesisBuilder {
         if !self.spentbook_nodes.is_empty() {
             // we only support a single mock spentbook section.  pubkeys must match.
             assert_eq!(
-                spentbook_node
-                    .key_manager
-                    .public_key_set()
-                    .unwrap()
-                    .public_key(),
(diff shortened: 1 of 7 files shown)


src/mock/genesis_builder.rs
```diff
@@ -56,15 +56,10 @@ impl GenesisBuilder {
...
-                    .key_manager
-                    .public_key_set()
-                    .unwrap()
-                    .public_key(),
+                spentbook_node.key_manager.public_key_set().public_key(),
                 self.spentbook_nodes[0]
                     .key_manager
                     .public_key_set()
-                    .unwrap()
                     .public_key()
```
davidrusu, line 63: You should be able to compare the `public_key_set()`'s themselves, shouldn't need to convert to `public_key()`

davidrusu (approved):
Nice change :+1:

PR #215, opened 2023-01-19
Remove RingCT

grumbach:
- Removes RingCT
- Keeps amount hiding with commitments and bulletproofs
- Keeps One Time Keys
- Replaces in-house crypto by higher level `blsttc` primitives like `PublicKey` (instead of manually doing it with `G1Affine`) or `PrivateKey` (instead of manually doing it with `Scalar`)


src/transaction/mod.rs
```diff
@@ -0,0 +1,64 @@
...
+            value,
+            blinding: BlindingFactor::random(&mut rng),
+        }
+    }
+
+    pub fn commit(&self, pc_gens: &PedersenGens) -> Commitment {
+        pc_gens
+            .commit(BlindingFactor::from(self.value), self.blinding)
+            .to_affine()
+    }
```
davidrusu, line 55: I don't think we gain much by taking the pc_gens as input. How about we just build the pc_gens inside commit to simplify the API.
```suggestion
    pub fn commit(&self) -> Commitment {
        PedersenGens::default()
            .commit(BlindingFactor::from(self.value), self.blinding)
            .to_affine()
    }
```

src/transaction/output.rs
```diff
@@ -0,0 +1,469 @@
...
+    for pk in public_keys.iter() {
+        msg.extend(pk.to_bytes().as_ref());
+    }
+    for r in input_commitments.iter() {
+        msg.extend(r.to_bytes().as_ref());
+    }
+    for o in output_proofs.iter() {
+        msg.extend(o.to_bytes());
+    }
+    msg
```
davidrusu, line 247: This is a paranoia thing but it's a good idea to put domain separators between the different field types to avoid situations where different messages may have the same byte encoding.

i.e. say for msg1, you had public_keys=[a,b,c] and commitments = [d]
and then msg2 where public_keys=[a,b] and commitments=[c,d]

If you look at the flattened byte encoding of the two messages, you'd see `<a><b><c><d>` for both messages.
Compare with domain seperators: 
msg1: `public_keys<a><b><c>commitments<d>`
msg2: `public_keys<a><b>commitments<c><d>`

```suggestion
    // Generate message to sign.
    let mut msg: Vec<u8> = Default::default();
    msg.extend("public_keys".as_bytes());
    for pk in public_keys.iter() {
        msg.extend(pk.to_bytes().as_ref());
    }
    msg.extend("commitments".as_bytes());
    for r in input_commitments.iter() {
        msg.extend(r.to_bytes().as_ref());
    }
    msg.extend("output_proofs".as_bytes());
    for o in output_proofs.iter() {
        msg.extend(o.to_bytes());
    }
    msg.extend("end".as_bytes());
    msg
```

src/transaction/output.rs
```diff
@@ -0,0 +1,469 @@
...
+}
+
+impl Ord for DbcTransaction {
+    fn cmp(&self, other: &Self) -> Ordering {
+        self.hash().cmp(&other.hash())
+    }
+}
+
+impl DbcTransaction {
+    pub fn to_bytes(&self) -> Vec<u8> {
```
davidrusu, line 300: Same as in https://github.com/maidsafe/sn_dbc/pull/215/files#r1093686688

davidrusu (dismissed):
Looks good! One point about how we are encoding messages for signing

davidrusu:
I think we should replace our `bls_bulletproofs` with the upstream bulletproofs since we no longer need them to be on the BLS curve, we'll get much better performance and it's audited. Follow up PR for sure though.

davidrusu (approved):
🚢

PR #218, opened 2023-02-08
Use bulletproofs instead of bls_bulletproofs

### LINKED ISSUE #216: Replace our `bls_bulletproofs` with the upstream bulletproofs
I think we should replace our `bls_bulletproofs` with the upstream bulletproofs since we no longer need them to be on the BLS curve, we'll get much better performance and it's audited. 

_Originally posted by @davidrusu in https://github.com/maidsafe/sn_dbc/issues/215#issuecomment-1412673059_

grumbach:
- Fixes #216 
- Significantly improves performance: 
    - `2096 ms` -> `234 ms` : `Benchmarking reissue split 1 to 100`
    -  `160s` -> `40s` : running all tests (on my PC)


davidrusu (approved):
😎

PR #221, opened 2023-02-10
Remove input's associated spent tx from the DBC struct

grumbach:
Remove input's associated spent transactions from the DBC struct:
- they where unused by `sn_dbc`
- they forced DBCs to include all input's Tx which doesn't apply to Genesis DBC
- broke `mint-repl` and benches as a result, thus breaking all `sn_dbc` operating examples
- removes extra verifications that don't prove anything apart from the fact that those spent transactions are indeed included in the DBC, the input's `SpentProof` is the one we want to check and do check.

Keeping them would involve a rather inelegant solution IMO:
- creating a fake Tx for genesis to tolerate this case
- creating a special case in the code just for the Genesis everywhere DBC checks are involved
- or maybe I didn't see something else?

### DIFF
--- a/benches/reissue.rs
+++ b/benches/reissue.rs
@@ -41,8 +41,8 @@ fn bench_reissue_1_to_100(c: &mut Criterion) {
         .build(&mut rng)
         .unwrap();
 
-    for (key_image, tx) in dbc_builder.inputs() {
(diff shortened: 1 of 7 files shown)


grumbach:
bors try

grumbach:
Putting this PR aside for now, as those spent transactions are used by the current `safe_network` implementation.

PR #233, opened 2023-02-27
feat(builder): add api for adding spentproof

oetyng:
This allows for spentproofs to be stored and returned in their aggregated form, as a builder can build spends from the full proofs as well and not only from shares.

### DIFF
--- a/src/builder.rs
+++ b/src/builder.rs
@@ -195,7 +195,7 @@ pub struct DbcBuilder {
     pub revealed_commitments: Vec<RevealedCommitment>,
     pub output_owner_map: OutputOwnerMap,
     pub revealed_tx: RevealedTransaction,
-
+    pub spent_proofs: HashSet<SpentProof>,
     pub spent_proof_shares: BTreeMap<PublicKey, HashSet<SpentProofShare>>,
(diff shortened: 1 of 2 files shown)


src/builder.rs
```diff
@@ -347,14 +354,18 @@ impl DbcBuilder {
...
             .iter()
             .map(|(key_image, shares)| {
                 SpentProof::try_from_proof_shares(*key_image, transaction_hash, shares.iter())
             })
             .collect::<Result<_>>()?;
 
+        self.spent_proofs.iter().for_each(|s| {
+            let _ = spent_proofs.insert(s.clone());
+        });
+
```
davidrusu, line 368: What do you think of using `chain` here instead of mutating?

```suggestion
        let spent_proofs: BTreeSet<SpentProof> = self
            .spent_proof_shares
            .iter()
            .map(|(key_image, shares)| {
                SpentProof::try_from_proof_shares(*key_image, transaction_hash, shares.iter())
            })
            .chain(self.spent_proofs.iter().cloned().map(|s| Ok(s)))
            .collect::<Result<_>>()?;

```

grumbach (approved):
LGTM

PR #236, opened 2023-03-14
Ground work for payments

grumbach:
- Adds public key and Commitment to dbc API

### DIFF
--- a/src/amount_secrets.rs
+++ b/src/amount_secrets.rs
@@ -7,12 +7,13 @@
 // permissions and limitations relating to use of the SAFE Network Software.
 
 use crate::transaction::{Amount, RevealedCommitment};
-use crate::{rand::RngCore, BlindingFactor, Error, Token};
+use crate::{rand::RngCore, BlindingFactor, Commitment, Error, Token};
 use blsttc::rand::CryptoRng;
 use blsttc::{
     Ciphertext, DecryptionShare, IntoFr, PublicKey, PublicKeySet, SecretKey, SecretKeySet,
     SecretKeyShare,
 };
+use bulletproofs::PedersenGens;
 use std::{collections::BTreeMap, convert::TryFrom};
 
 #[cfg(feature = "serde")]
@@ -43,6 +44,11 @@ impl AmountSecrets {
     /// blinding factor getter
     pub fn blinding_factor(&self) -> BlindingFactor {
         self.0.blinding
+    }
+
+    /// commitment
+    pub fn commitment(&self) -> Commitment {
+        self.0.commit(&PedersenGens::default())
     }
 
     /// Convert to bytes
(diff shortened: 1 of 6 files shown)


src/amount_secrets.rs
```diff
@@ -45,6 +46,11 @@ impl AmountSecrets {
         self.0.blinding
     }
 
+    /// commitment
```
oetyng, line 49: Comments with just saying the name of a fn is not doing much :)

src/dbc.rs
```diff
@@ -66,9 +66,15 @@ use crate::{
 #[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
 #[derive(Debug, Clone, Eq, PartialEq)]
 pub struct Dbc {
+    /// Encrypted information for and about this DBC's owner
```
oetyng, line 69: Great that you filled in these missing comments!

Definitely helps to have the main struct members described!

src/dbc.rs
```diff
@@ -66,9 +66,15 @@ use crate::{
...
     pub content: DbcContent,
+    /// The public key (owner) of this DBC, used for SpentProofs
+    pub public_key: PublicKey,
```
oetyng, line 72: Is this not accessible through the `DbcContent`?

src/dbc.rs
```diff
@@ -66,9 +66,15 @@ use crate::{
...
+    /// The transaction where this DBC was created
     pub transaction: DbcTransaction,
+    /// The transaction's input's SpentProofs
     pub spent_proofs: BTreeSet<SpentProof>,
```
oetyng, line 76: I would even consider renaming these to:
`input_spent_proofs` and `input_spent_txs` for additional clarity.

src/dbc.rs
```diff
@@ -390,7 +411,37 @@ pub(crate) mod tests {
 
     #[test]
     fn from_hex_should_deserialize_a_hex_encoded_string_to_a_dbc() -> Result<(), Error> {
-        let dbc = Dbc::from_hex(DBC_WITH_1_530_000_000)?;
```
oetyng, line 393: Is this replacement really necessary?
Cant the public key just be gotten from this preprocessed dbc's dbc content?

src/dbc.rs
grumbach: Only accessible by the DBC owner through the encrypted derivation index (as it used to be a secret). 
This whole thing should be reworked and maybe we should drop that part altogether later. 
But for now, making it public is important for usability.

src/dbc.rs
grumbach: Not necessary. But this will save us 15 minutes every time we change the DBC struct again.

oetyng (approved):

PR #238, opened 2023-03-22
refactor!: clarify use of blinded and revealed amount

oetyng:
This makes it much clearer what we in the domain are handling; Blinded and revealed amounts.

The concept "commitment" comes from `Pedersen Commitment`, which is the tech we use to implement _blinded amounts_.
What we were doing was to try reflect an impl detail, by calling it "commitment", thus totally missing the aspect of it being a _blinded amount_, which was and is the important aspect in our domain. This is now named `BlindedAmount`.

To match that and move towards more stringency, the `AmountSecrets` wrapper around `RevealedAmount` is removed, which removes yet another obscure concept that was causing cognital overhead.


davidrusu (approved):
Looks good

PR #256, opened 2023-04-10
chore: expose public address in the api

oetyng:
This is used by a wallet, should have been exposed already but was missed.

A few remainder renames in docs and internal fns is done here as well.


grumbach (approved):
LGTM

PR #261, opened 2023-04-11
refactor!: include spend src tx hash and simplify builder api

oetyng:
The `spend src tx hash` is the hash of the tx where the dbc to be spent, was created. In other words: that src tx shall contain this spend as an output.

Also extends misc apis.

### DIFF
--- a/src/builder.rs
+++ b/src/builder.rs
@@ -12,9 +12,9 @@ use std::collections::{BTreeMap, BTreeSet};
 use crate::{
     dbc_id::DbcIdSource,
     transaction::{
-        DbcTransaction, Output, RevealedAmount, RevealedInput, RevealedOutput, RevealedTx,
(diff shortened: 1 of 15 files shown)


src/builder.rs
```diff
@@ -65,24 +78,24 @@ impl TransactionBuilder {
...
-
-    /// Add an input given a list of (DerivedKey, RevealedAmount).
-    pub fn add_inputs_by_secrets(mut self, secrets: Vec<(DerivedKey, RevealedAmount)>) -> Self {
-        for (derived_key, revealed_amount) in secrets.into_iter() {
-            self = self.add_input_by_secrets(derived_key, revealed_amount);
-        }
-        self
-    }
+    // /// Add an input given a DerivedKey and a RevealedAmount.
+    // pub fn add_input_by_secrets(
```
grumbach, line 82: Do we need to keep these commented lines?

grumbach (approved):
LGTM

PR #265, opened 2023-04-12
refactor!: have dbcs take derived key

oetyng:
Have dbcs take derived key. This makes more sense in upper layer (bearer scenario for testing)

Correct fn name, prefixing with `as` is usually done consuming self, which we don't do.

Remove commented out.


grumbach (approved):
LGTM

PR #278, opened 2023-06-14
feat: include parent tx in spend

grumbach:


RolandSherwin (approved):
LGTM! The `parent_tx` is used for validations in `safe_network`, this would simplify things there.

PR #282, opened 2023-06-29
feat: adding a fee output type

bochaco:


src/builder.rs
```diff
@@ -113,7 +120,12 @@ impl TransactionBuilder {
 
     /// Get sum of output Tokens.
     pub fn outputs_tokens_sum(&self) -> Token {
-        let amount = self.outputs.iter().map(|o| o.token.as_nano()).sum();
+        let amount = self
+            .outputs
+            .iter()
+            .map(|o| o.token.as_nano())
+            .chain(std::iter::once(self.fee.token.as_nano()))
+            .sum();
```
RolandSherwin, line 128: Perhaps we should have a `checked_add` here to error out on an overflow? Maybe I can work on it.

RolandSherwin (approved):

PR #291, opened 2025-07-04
fixed forum link because of domain name change

rid-dim:
just a fix for the DBC "writeups about the technology" that have been linked but the domain name of the forum changed by now


dirvine (approved):
Perfect, thank you for this

